PR #8, opened 2023-09-25
Add input/output labels and bottom result warning

julien-nc:
image

Not sure about the labels, for the input, would something like "Text" be better?


jancborchardt (commented):
How about "Text to summarize" und "Summary"? :)

julien-nc:
> How about "Text to summarize" und "Summary"? :)

Not possible now as we can't make sure the provider name is a verb or an action. We could do something like that once the TextProcessing API has been improved and forces the providers to define those labels.

Any idea of good generic labels?

nimishavijay (commented):
Nice! :) 
- [x] We could move the "generated by AI" warning into an NcNoteCard in the warning style to make it more attention grabbing? or is it too much? @jancborchardt 
- [x] About the warning wording: "This output was generated by AI. Make sure to double-check and adjust." @jancborchardt Not sure about this also tbh, do we use "please" or "make sure" much?
- [x] "Input" and "Result" seem good to me until we are able to have more specific words :)

jancborchardt (commented):
Looks really nice @julien-nc! :) Super small detail: Could the info box always span the exact same width of the textareas? Currently it has a bit of a gap to the right.

julien-nc:
@jancborchardt Sure
image

nimishavijay (approved):
🚀

PR #18, opened 2023-12-07
Integration improvements

julien-nc:
refs https://github.com/nextcloud/text/issues/4686

* [x] The main promise always resolves the task with its current status
* [x] Assistant will offer a flag to close the modal after a successful sync task
* [x] There will be a list of action buttons that can be provided

https://github.com/nextcloud/assistant/assets/11291457/5d579c0b-7ee5-4368-97ef-78ff519ff46d

``` js
const params = {
	appId: 'plopApp',
	identifier: 'idid',
	input: 'give me a short summary of a simple settings section about GitHub',
	closeOnResult: false,
	actionButtons: [
		{
			label: 'Put in field 1',
			title: 'Title 1',
			type: 'warning',
			iconSvg: cogSvg,
			onClick: (task) => { this.field1 = task.output },
		},
		{
			label: 'Put in field 2',
			title: 'Title 2',
			onClick: (task) => { this.field2 = task.output },
		},
	],
}
OCA.TPAssistant.openAssistantForm(params)
```

Is it enough to get the task output as parameter of button callbacks? I could refactor a bit and make it possible to get the task itself if necessary.

### DIFF
--- a/src/assistant.js
+++ b/src/assistant.js
@@ -11,7 +11,21 @@ __webpack_public_path__ = linkTo('assistant', 'js/') // eslint-disable-line
  *  appId: 'my_app_id',
  *  identifier: 'my task identifier',
  *  taskType: 'OCP\\TextProcessing\\FreePromptTaskType',
- *  input: 'count to 3'
+ *  input: 'count to 3',
+ *  actionButtons: [
+ *    {
+ *      label: 'Label 1',
+ *      title: 'Title 1',
+ *      type: 'warning',
+ *      iconSvg: cogSvg,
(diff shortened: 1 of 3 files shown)


julien-nc:
Done.

One little detail: If the user edits the result before pressing the action button, the value of the text area is passed to the action button, not the value that we really got in the task. I thought it would make sense as users might want to adjust the result in the Assistant UI before pressing a button.

julien-nc:
Oh and the code example has been adjusted.

juliushaertl (approved):
I won't get to testing/integrating this now, but code looks good 👍

PR #20, opened 2023-12-18
Add image generation smart picker

MB-Finski:
Migrates the functionality of text2image_helper under assistant.

Features:
- Text2image smart picker
- Local storage of images with customizable TTL (can be set by admin). Each view of a generation resets the TTL timestamp.
- Image generation pages can be shared publicly; all relevant endpoints are either bruteforsce protected or rate limited
- Dynamic UI based on whether the task was scheduled or run immediately; user can ask for a notification of a generation being complete if it was scheduled to run in the background.

The focus has been to refactor as little of the code as possible.

Some points that could use improvement:
- The term generation (referring potentially to multiple images) is a bit confusing; maybe 'group' would be more descriptive?
- Currently does not import images or migrate the db from text2image_helper. Should we do that?


src/reference.js
```diff
@@ -0,0 +1,41 @@
...
+
+import { registerWidget, registerCustomPickerElement, NcCustomPickerRenderResult } from '@nextcloud/vue/dist/Components/NcRichText.js'
+
+import { linkTo } from '@nextcloud/router'
+import { getRequestToken } from '@nextcloud/auth'
+
+__webpack_nonce__ = btoa(getRequestToken()) // eslint-disable-line
+__webpack_public_path__ = linkTo('assistant', 'js/') // eslint-disable-line
+
+registerCustomPickerElement('assistant', async (el, { providerId, accessible }) => {
```
julien-nc, line 12:
```suggestion
registerCustomPickerElement('assistant_image_generation', async (el, { providerId, accessible }) => {
```

webpack.js
```diff
@@ -15,6 +15,8 @@ webpackConfig.stats = {
 
 const appId = 'assistant'
 webpackConfig.entry = {
+	generationPage: { import: path.join(__dirname, 'src', 'imageGenerationPage.js'), filename: appId + '-imageGenerationPage.js' },
```
julien-nc, line 18:
```suggestion
	imageGenerationPage: { import: path.join(__dirname, 'src', 'imageGenerationPage.js'), filename: appId + '-imageGenerationPage.js' },
```
A little bit of nitpicking :grin:

webpack.js
```diff
@@ -15,6 +15,8 @@ webpackConfig.stats = {
...
+	reference: { import: path.join(__dirname, 'src', 'reference.js'), filename: appId + '-reference.js' },
```
julien-nc, line 19: As we'll add other reference stuff and they will probably have their own script each, let's name this one `imageGenerationReference.js`

julien-nc (changes requested):
Noice!
I'll make a second review after having tried it in action.

lib/Reference/Text2ImageReferenceProvider.php
```diff
@@ -0,0 +1,154 @@
...
+		private ImageGenerationMapper $imageGenerationMapper,
+		private ?string $userId
+	) {
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getId(): string {
+		return 'assistant';
```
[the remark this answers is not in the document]
MB-Finski: Yeah, I realized I need to do this while porting gptfreeprompt just now!

lib/Service/AssistantService.php
```diff
@@ -36,14 +37,20 @@ public function sendNotification(Task $task, ?string $target = null, ?string $ac
...
+			$params['taskType'] = Application::TASK_TYPE_TEXT_GEN;
+			$params['textTaskTypeClass'] = $task->getType();
+		}
+
 		$status = $task->getStatus();
 		$subject = $status === Task::STATUS_SUCCESSFUL
 			? 'success'
 			: 'failure';
-		$objectType = ($task->getAppId() === Application::APP_ID || $target === null)
+		$objectType = ($task->getAppId() === Application::APP_ID && $target === null)
```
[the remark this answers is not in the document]
MB-Finski: I did this change since the text to image tasks need a different target from the text processing tasks but I guess we could also resolve that target in the notifier, as well. Only difference being that we also need to pass the task identifier to resolve the link.

src/components/Text2Image/Text2ImageDisplay.vue
```diff
@@ -0,0 +1,436 @@
...
+
+	.image-wrapper {
+		display: flex;
+		flex-direction: column;
+		position: relative;
+		max-width: 100%;
+		height: 100%;
+		margin-top: 12px;
+		filter: grayscale(100%);
+		transition: filter 0.2s ease-in-out;
```
julien-nc, line 349: This does not work on my side and the images stay gray everywhere they get displayed (preview in the picker, link preview widget).

MB-Finski: Hmm.. That's unfortunate. On Firefox, or?

src/views/Text2Image/Text2ImageCustomPickerElement.vue
```diff
@@ -0,0 +1,427 @@
+<template>
+	<div class="text2image-picker-content-wrapper">
+		<div class="text2image-picker-content">
+			<h2>
+				{{ t('assistant', 'AI Image Generation') }}
+			</h2>
+			<div class="input-wrapper">
+				<NcTextField ref="text2image-search-input"
+					:value.sync="query"
+					:label="inputPlaceholder"
```
julien-nc, line 10:
```suggestion
					:label="inputPlaceholder"
					placeholder=""
```
In NcInputField/NcTextField, the label is used as placeholder if no placeholder is set. Let's use either an empty one or a query example.

src/components/Text2Image/Text2ImageDisplay.vue
MB-Finski: Note to self: filter should be none for the default state.

lib/Service/AssistantService.php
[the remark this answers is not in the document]
MB-Finski: Note to self: $target === null is the correct condition here.

lib/Service/AssistantService.php
```diff
@@ -8,6 +8,7 @@
 use OCP\PreConditionNotMetException;
 use OCP\TextProcessing\IManager as ITextProcessingManager;
 use OCP\TextProcessing\Task;
+use OCP\TextToImage\Task as TextToImageTask;
```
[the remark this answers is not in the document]
MB-Finski: Hmm.. To make the code more readable in the future, it might also make sense to rename the subsequent methods? Also the controller methods?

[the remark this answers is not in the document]
MB-Finski: In the latest commit I only renamed the methods in the service, but....

MB-Finski:
Hmm... Specifying 'cursor: pointer' didn't work for the check-box issue. I'll have to look into it a bit more.

MB-Finski:
BTW, one improvement would be that the generation page could be changed to be a modal like it is in other cases with assistant.

lib/Service/AssistantService.php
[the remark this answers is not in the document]
julien-nc: Then let's rename the controller methods as well (don't forget to adjust routes.php).

lib/Service/AssistantService.php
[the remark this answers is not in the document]
julien-nc: :+1:

lib/Migration/Version010003Date20231208083112.php
```diff
@@ -0,0 +1,133 @@
...
+			$table->addColumn('user_id', Types::STRING, [
+				'notnull' => true,
+				'length' => 64,
+			]);
+			$table->addColumn('value', Types::STRING, [
+				'notnull' => true,
+				'length' => 1000,
+			]);
+			$table->addColumn('timestamp', Types::BIGINT, [
+				'notnull' => true,
```
julien-nc, line 50:
```suggestion
				'notnull' => true,
				'unsigned' => true,
```
For all the timestamps.

julien-nc (changes requested):
Small thing and let's get this in.

julien-nc:
> BTW, one improvement would be that the generation page could be changed to be a modal like it is in other cases with assistant.

Let's keep that in mind with #21.

MB-Finski:
Just a moment, I'll still need to fix the link generation for the copied link!

julien-nc:
Take your time, don't forget to rename the controller methods.

MB-Finski:
Hmm.. Now I think we're close!

julien-nc (approved):
:rocket:

PR #22, opened 2023-12-22
Add a free prompt smart picker

MB-Finski:
- Adds:
  - A basic smart picker front-end for the free prompt task type

- Notes:
  - An obvious refactoring target would be to move the front end to use the end points provided by the assistant core service, although they do not match the current functionalities 1 to 1.


MB-Finski:
Why that "[Add image generation smart picker](https://github.com/nextcloud/assistant/pull/22/commits/f011612ca9ca6458fd200e05b5b040a08fc854b5)" commit has come to haunt me here, I have no idea. Probably something funny happened with the multiple rebases that I did... :/

lib/Migration/Version010003Date20231208083112.php
```diff
@@ -122,6 +122,29 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			$table->addIndex(['image_gen_id'], 'assistant_i_stale_gens_id');
 		}
 
+		if (!$schema->hasTable('assistant_t_prompts')) {
```
julien-nc, line 125: Small detail for this migration. We usually have a boolean flag named `$schemaChanged` which is initialized to false and becomes true if there actually are changes to the schema. At the end of `changeSchema()` we return the schema if there were changed or null if not.

Small optimization to avoid returning a schema when there were no changes.

lib/Settings/Personal.php
```diff
@@ -24,7 +25,10 @@ public function __construct(
 	 * @return TemplateResponse
 	 */
 	public function getForm(): TemplateResponse {
-		$textProcessingAvailable = $this->textProcessingManager->hasProviders() ?
+		$textProcessingAvailable = $this->textProcessingManager->hasProviders();
+		$freePromptTaskTypeAvailable = in_array(FreePromptTaskType::class, $this->textProcessingManager->getAvailableTaskTypes());
+
+		$assistantAvailable = $textProcessingAvailable ?
```
julien-nc, line 31:
```suggestion
		$assistantAvailable = $textProcessingAvailable && $this->config->getAppValue(Application::APP_ID, 'assistant_enabled', '1') === '1'
```
Right?

lib/Settings/Personal.php
```diff
@@ -33,12 +37,19 @@ public function getForm(): TemplateResponse {
 			$this->config->getAppValue(Application::APP_ID, 'text_to_image_picker_enabled', '1') === '1' :
 			false;
 		$textToImagePickerEnabled = $this->config->getUserValue($this->userId, Application::APP_ID, 'text_to_image_picker_enabled', '1') === '1';
+
+		$freePromptPickerAvailable = $freePromptTaskTypeAvailable ?
+			$this->config->getAppValue(Application::APP_ID, 'free_prompt_picker_enabled', '1') === '1' :
+			false;
```
julien-nc, line 43:
```suggestion
		$freePromptPickerAvailable = $freePromptTaskTypeAvailable && $this->config->getAppValue(Application::APP_ID, 'free_prompt_picker_enabled', '1') === '1';
```

lib/Settings/Personal.php
```diff
@@ -33,12 +37,19 @@ public function getForm(): TemplateResponse {
...
+		$freePromptPickerEnabled = $this->config->getUserValue($this->userId, Application::APP_ID, 'free_prompt_picker_enabled', '1') === '1';
 		
 		$userConfig = [
-			'text_processing_available' => $textProcessingAvailable,
+			'assistant_available' => $assistantAvailable,
```
julien-nc, line 47: Mistake or you don't want to name this `text_processing_available` anymore?

It does not really matter in the end as long as its meaning is explicit and clear.

julien-nc (changes requested):
Small remarks but all good :+1:

lib/Settings/Personal.php
MB-Finski: Here I thought that the difference is that in user settings it may be that text processing is available but that the admin has disabled the feature. That's why I thought it would make more sense here to name it assistant_available (since it may just be disabled by the admin).

MB-Finski: Hmm. Should it be top_bar_assistant_available/enabled here instead?

MB-Finski:
Pushed some fixes. I also have the stt_helper migration ready for once this is ready to be merged.

lib/Settings/Personal.php
julien-nc: Yes, whatever makes it very explicit on what it means and what it has an impact on

julien-nc (approved):

PR #27, opened 2024-01-11
Add stt smart picker

MB-Finski:
This pr imports the stt smart picker functionality from the stt_helper.

Again I tried to keep the refactoring to a minimum, although, I had some hesitation about doing some refactoring for the notification flow (since you cannot use the tasks like they are used for text processing notifications).


lib/Service/SpeechToText/SpeechToTextService.php
```diff
@@ -0,0 +1,199 @@
...
+		} catch (\OCP\Db\Exception $e) {
+			$this->logger->error('Failed to save transcript in DB: ' . $e->getMessage());
+			$success = false;
+		}
+
+		$params = [
+			'appId' => Application::APP_ID,
+			'taskType' => Application::TASK_TYPE_SPEECH_TO_TEXT,
+			'result' => $result,
+			'target' => $this->urlGenerator->linkToRouteAbsolute(Application::APP_ID . '.SpeechToText.getResultPage', ['id' => $id])
```
julien-nc, line 185: `$id` is not defined in case of error. Let's set it to 0 in the previous catch block (catching `\OCP\Db\Exception`).

src/components/AssistantPlainTextModal.vue
```diff
@@ -0,0 +1,258 @@
...
+		.icon {
+			margin-right: 8px;
+		}
+	}
+
+	.editable-output {
+		width: 100%;
+		display: flex;
+		flex-direction: column;
+		align-items: start;
```
julien-nc, line 219:
```suggestion
```
This is causing NcRichContenteditable to shrink to the width of the text content.

src/components/AssistantPlainTextModal.vue
```diff
@@ -0,0 +1,258 @@
...
+import Vue from 'vue'
+
+import { showError } from '@nextcloud/dialogs'
+
+import { TASK_TYPES } from '../constants.js'
+
+Vue.use(VueClipboard)
+
+export default {
+	name: 'AssistantPlainTextModal',
```
julien-nc, line 90: Is this component going to be used in other contexts than STT? If not, we could rename it to `SttResultModal`.

julien-nc (changes requested):
Nice!
Whenever we add a migration step, it's better to bump the app version number so we make sure the step is triggered when people test the PR (i was confused for a moment).
A few minor changes and we're good to go.

src/components/AssistantPlainTextModal.vue
MB-Finski: I must admit, I have no other definite use case for a plain text result page yet but I ventured a guess here that we might have those in the future (for example, notifications where we want to hide which text processing task type was responsible for the result) . Although, it might also be that such features would be otherwise accommodated in the notification logic?

julien-nc (approved):

src/components/AssistantPlainTextModal.vue
julien-nc: Anyway, not a big deal. Rename it if you want, merge if you don't :grin:

PR #29, opened 2024-01-12
fix(i18n): Fixed typo

rakekniven:


julien-nc:
Oh, didn't spot that. Thanks @rakekniven 
@MB-Finski It's best to not translate the exception messages as it's sometimes hard to translate and admins are used to english terminology anyway.

PR #34, opened 2024-01-29
Add rudimentary copywriter functionality, implement assistant metatask wrapper

MB-Finski:
This pr implements rudimentary copywriter functionality for assistant.

- Uses a free prompt provider in the background
- Implements an assistant meta-task wrapper to differentiate between plain free prompt tasks and custom internal task types
- Generalizes the wrapper for all current task modalities to unify the notification system.
- Implements a text file picker for all assistant task types in the front-end. Currently supports ODT, DOC, DOCX, RTF, MD and plain text by using PhpOffice/PhpWord and parsedown to parse the files at the backend.

Known issues:
- Some psalm issues still remain
- The UI is not ideal but possibly "enough for an MVP".


MB-Finski:
Oh, and task "modality" has not yet been renamed: I'm open for suggestions!

julien-nc:
Suggestions:
* family
* category
* group
* taskOcpNamespace
* feature

appinfo/info.xml
```diff
@@ -49,9 +49,11 @@ include text processing providers to:
 	<screenshot>https://github.com/nextcloud/assistant/raw/main/img/screenshot3.jpg</screenshot>
 	<background-jobs>
 		<job>OCA\TpAssistant\Cron\CleanupImageGenerations</job>
+		<job>OCA\TpAssistant\Cron\CleanupAssistantTasks</job>
 	</background-jobs>
 	<commands>
 		<command>OCA\TpAssistant\Command\CleanupImageGenerations</command>
+		<command>OCA\TpAssistant\Command\CleanupAssistantTasks</command>
```
marcelklehr, line 56: Everytime I read TpAssistant, I think of this:

image

lib/Command/CleanupAssistantTasks.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+// SPDX-FileCopyrightText: Sami Finnilä <sami.finnila@nextcloud.com>
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+namespace OCA\TpAssistant\Command;
+
+use Exception;
+use OC\Core\Command\Base;
```
marcelklehr, line 9: Is there no OCP equivalent for this?

marcelklehr:
I feel like there are some files missing?

lib/Command/CleanupAssistantTasks.php
julien-nc: Unfortunately not. Even spreed is using this OC class. https://github.com/nextcloud/spreed/blob/main/lib/Command/Command/ListCommand.php#L26

julien-nc:
> I feel like there are some files missing?

@marcelklehr Which ones?

marcelklehr:
>> 
>>  I feel like there are some files missing?
>
> @marcelklehr Which ones?

Nevermind, may have been a Github glitch

lib/Controller/AssistantController.php
```diff
@@ -33,9 +34,10 @@ public function __construct(
 	 */
 	#[NoAdminRequired]
 	#[NoCSRFRequired]
-	#[BruteForceProtection(action: 'taskResultPage')]
+	#[BruteForceProtection(action: 'taskResults')]
```
marcelklehr, line 37: I continue to be flabbergasted by BruteforceProtection.. :thinking: 

[Our docs](https://docs.nextcloud.com/server/latest/developer_manual/basics/controllers.html#brute-force-protection) seem to indicate we need to pass in `['action' => 'taskResults']` to throttle?

lib/Db/Task.php
```diff
@@ -0,0 +1,108 @@
...
+			'task_type' => $this->taskType,
+			'timestamp' => $this->timestamp,
+			'status' => $this->status,
+			'category' => $this->category,
+			'indentifer' => $this->indentifer,
+		];
+	}
+
+	#[\ReturnTypeWillChange]
+	public function jsonSerializeCc() {
```
marcelklehr, line 86: What's this about? :)

lib/Controller/AssistantController.php
```diff
@@ -47,45 +49,105 @@ public function getTextProcessingTaskResultPage(int $taskId): TemplateResponse {
...
 		return new TemplateResponse(Application::APP_ID, 'taskResultPage');
 	}
 
 	/**
-	 * @param string $input
+	 * @param int $taskId
+	 * @return DataResponse
+	 */
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
```
marcelklehr, line 61: Any particular reason to disable CSRF here?

lib/Service/AssistantService.php
```diff
@@ -66,50 +103,300 @@ public function sendNotification(TextProcessingTask|TextToImageTask $task, ?stri
...
+
+			if($ocpTask->getStatus() !== $task->getStatus()) {
+				$task->setStatus($ocpTask->getStatus());
+				$task->setOutput($ocpTask->getOutput());
+				$task = $this->taskMapper->update($task);
+			}
+		} catch (NotFoundException $e) {
+			// Ocp task not found, so we can't update the status
+		} catch (\Exception | \Throwable $e) {
+			// Something else went wrong, so we can't update the status
```
marcelklehr, line 173: Generic catch(\Throwable|\Exception) is dangerous. Can we avoid this here?

lib/Service/AssistantService.php
```diff
@@ -66,50 +103,300 @@ public function sendNotification(TextProcessingTask|TextToImageTask $task, ?stri
...
+		
+
+		try {
+			$mimeType = $userFolder->get($filePath)->getMimeType();
+		} catch (NotFoundException $e) {
+			throw new \Exception('File not found.');
+		}
+
+		try {
+			$contents = $userFolder->get($filePath)->getContent();
```
marcelklehr, line 306: Psalm complains because this may be a folder

lib/Controller/AssistantController.php
julien-nc: I think the doc is weird/wrong. What you pass to `throttle` is the additional metadata, not the action again. The action of the created entries in the bruteforce_attempts table is determined by the method attribute `action` parameter.

You can actually pass anything you want as metadata so you know more about why something was throttled.

lib/Service/FreePrompt/FreePromptService.php
```diff
@@ -26,18 +27,17 @@ public function __construct(
...
 
 	/*
 	 * @param string $prompt
-	 * @param int $nResults
-	 * @param bool $showPrompt
 	 * @return string
 	 * @throws Exception
 	 */
-	public function processPrompt(string $prompt, int $nResults): string {
+	public function processPrompt(string $prompt): string {
```
marcelklehr, line 40: Seems like this is duplicating the text processing task logic in AssistantService, no?

marcelklehr: Instead of having process and schedule methods all over the place, I could imagine we could make use of the visitor pattern with a RunVisitor, a ScheduleVisitor, a RunOrScheduleVisitor and perhaps a NotifyVisitor or similar that each visit a Task

lib/Controller/AssistantController.php
marcelklehr: ah ok

lib/Service/FreePrompt/FreePromptService.php
MB-Finski: Yeah, I absolutely agree! I guess the question is if such refactoring should be a part of this pr?

lib/Db/Task.php
MB-Finski: Ha! Well it's nicer to have camel case in the front-end since we pass the serialized tasks directly there. It's just for aesthetics, basically...

lib/Controller/AssistantController.php
julien-nc: Oh I see, if there is an `action` key in the metadata array, it is used to know to which bruteforce this throttling is related (in case there are multiple protections to one method). But one can actually add whatever other keys in metadata.

In this case there is only one bruteforce protection method attr so we don't need to specify the action in the metadata.

lib/Db/Task.php
marcelklehr: Ah, I get it. CC is for CamelCase :see_no_evil:

lib/Controller/AssistantController.php
MB-Finski: Yeah, shouldn't be disabled here!

lib/Service/FreePrompt/FreePromptService.php
marcelklehr: Yeah, let's not make it part of this PR

lib/Controller/FreePromptController.php
```diff
@@ -33,9 +33,9 @@ public function __construct(
 	 */
 	#[NoAdminRequired]
 	#[NoCSRFRequired]
-	public function processPrompt(string $prompt, int $nResults = 1): DataResponse {
+	public function processPrompt(string $prompt): DataResponse {
 		try {
-			$result = $this->freePromptService->processPrompt($prompt, $nResults);
+			$result = $this->freePromptService->processPrompt($prompt);
```
julien-nc, line 38: The logic of checking the userId should be moved from `FreePromptService` to `FreePromptController`. The controller checks if it's valid (is null in our case) and, if so, calls the service method passing the userId as param, if not, returns an error response.

The `$userId` param of service methods can have the `string` type (not `?string`) if you know this action only makes sense for an authenticated user and the userId has been checked in the controller. This applies to all controllers and services.

lib/Listener/SpeechToText/SpeechToTextResultListener.php
```diff
@@ -49,20 +55,69 @@ public function handle(Event $event): void {
...
+			// Find a matching etag:
+			$etag = $file->getEtag();
+			$assistantTask = null;
+			foreach ($tasks as $task) {
+				$taskEtag = $task->getInputsAsArray()['eTag'];
+				if ($taskEtag === $etag) {
+					$assistantTask = $task;
+					break;
+				}
+			}
```
julien-nc, line 71: Why would there be multiple tasks with the same ocp_task_id?

julien-nc (changes requested):
Huge work :+1: 
I reviewed half of the PR approximately. The rest is coming this afternoon.

lib/Listener/SpeechToText/SpeechToTextResultListener.php
MB-Finski: That's an issue specific to the speech-to-text API; it doesn't have task ids so the only way to resolve the notification event back to the assistant task is by using the file id which is substituted for the ocpTaskId in this case. Then it's possible that you could have multiple tasks for the same file id (and possibly multiple different file versions)..?

lib/Controller/FreePromptController.php
MB-Finski: Ah, yeah, I forgot to do that!

lib/Controller/AssistantController.php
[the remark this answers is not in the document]
MB-Finski: I guess I was thinking this from the point of view of an attacker. With two different actions I could get twice the number of attempts with different actions. It may well be that I've misunderstood the functioning of the brute force protection as well...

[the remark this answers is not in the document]
MB-Finski: ... But do we need brute force protection here anymore since the service method only returns the task if it's owned by the user?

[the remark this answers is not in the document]
julien-nc: Yeah i thought the same, we don't need it if it's not a public endpoint. Let's remove it.

lib/Service/AssistantService.php
```diff
@@ -66,50 +104,308 @@ public function sendNotification(TextProcessingTask|TextToImageTask $task, ?stri
...
+		try {
+			$file = $userFolder->get($filePath);
+			if ($file instanceof File) {
+				$contents = $file->getContent();
+			} else {
+				throw new \Exception('File is not a file.');
+			}
+		} catch (NotFoundException | LockedException | GenericFileException | NotPermittedException $e) {
+			throw new \Exception('File not found or could not be accessed.');
+		}
```
julien-nc, line 318: You could first get the file, then check it's an instance of `OCP\Files\File`, then get the mime type. I mean let's avoid getting the file twice.

lib/Service/FreePrompt/FreePromptService.php
```diff
@@ -24,78 +25,77 @@ public function __construct(
...
 	/*
 	 * @param string $prompt
-	 * @param int $nResults
-	 * @param bool $showPrompt
+	 * @param string $userId
 	 * @return string
 	 * @throws Exception
 	 */
-	public function processPrompt(string $prompt, int $nResults): string {
+	public function processPrompt(string $prompt, $userId): string {
```
julien-nc, line 40:
```suggestion
	public function processPrompt(string $prompt, string $userId): string {
```

julien-nc (changes requested):

lib/Service/AssistantService.php
```diff
@@ -66,50 +104,308 @@ public function sendNotification(TextProcessingTask|TextToImageTask $task, ?stri
...
+		try {
+			$ocpTask = $this->textProcessingManager->getTask($task->getOcpTaskId());
+
+			if($ocpTask->getStatus() !== $task->getStatus()) {
+				$task->setStatus($ocpTask->getStatus());
+				$task->setOutput($ocpTask->getOutput());
+				$task = $this->taskMapper->update($task);
+			}
+		} catch (NotFoundException $e) {
+			// Ocp task not found, so we can't update the status
```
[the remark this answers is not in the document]
MB-Finski: It's conceivable that an assistant meta task is stored for longer than the corresponding ocp task. This would be the case, for example, with generated images, if we unify the routes and front-end for all task categories at some point in the future. That's why I initially considered the generic catching of all exceptions acceptable here, since the primary task of the method has already been completed (although it definitely makes sense to do some debug logging).

lib/Service/AssistantService.php
```diff
@@ -66,50 +104,308 @@ public function sendNotification(TextProcessingTask|TextToImageTask $task, ?stri
...
-	 * @return TextProcessingTask
+	 * @return Task
 	 * @throws PreConditionNotMetException
+	 * @throws \Exception
 	 */
-	public function runTextProcessingTask(string $type, string $input, string $appId, ?string $userId, string $identifier): TextProcessingTask {
-		$task = new TextProcessingTask($type, $input, $appId, $userId, $identifier);
-		$this->textProcessingManager->runTask($task);
-		return $task;
+	public function runTextProcessingTask(string $type, array $inputs, string $appId, ?string $userId, string $identifier): Task {
```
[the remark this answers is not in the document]
MB-Finski: Oh, that's true! I always thought of copywriter as just another text processing task but it might as well be considered another task category...

lib/Service/AssistantService.php
MB-Finski: Makes sense!

lib/Controller/AssistantController.php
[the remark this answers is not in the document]
MB-Finski: Ok, I also blocked the scheduling/running of tasks with "userId === null" while at it.

julien-nc (approved):

PR #38, opened 2024-02-02
First round of design improvements

julien-nc:
* Rename "free prompt" to "generate text"
* Rename "speech to text" to "AI audio transcription"
* Simplify design of STT recorder
* Fix accessibility of some buttons
* Fix keyboard navigation in task type selector
* Fix style of image generation preview
* Warn about empty result or if it's identical to the input
* Change description of the free prompt task type
* Fix z-index of a modal close button
* Fix updating input fields when file content is loaded


MB-Finski (approved):
Looks and feels much better! 

As a semi-related thought; maybe adding "align-items: center;" in the Text2ImageCustomPickerElement .image-preview would make sense? Currently the image display is centered everywhere else but not in the picker itself.

BTW the bug related to output resetting still exists in gptfreeprompt but is not reproducible in assistant; like we discussed the more up-to-date libs probably explain that.

julien-nc:
@MB-Finski Good point. I went all the way and unified the style of the image previews everywhere.
The width of Text2ImageDisplay is not defined in the component itself anymore but rather at each place it is used.
There's now a border around the preview in the preview page just like in other places.
The preview container in the picker is now full width so it aligns with the rest.
Wdyt of all that?

PR #42, opened 2024-02-08
Move STT and image generation to the assistant UI

julien-nc:
* Fix image generation result page for non-authenticated users
* Add an STT task type in the assistant UI and implement a specific input form
* Add an image generation task type in the assistant UI and implement a specific output field
* Support STT and image generation in the assistant task result page
* Use the assistant result page as target for STT and image generation notifications
* Allow to schedule (instead of runOrSchedule) image generation tasks (when canceling a sync task with the "run in the background" button)
* Harmonize image generation workflow with the rest: don't show the result widget if it has been scheduled, set notifyReady to true automatically
* Use `@nextcloud/moment` instead of `humanize-duration` for time estimation calculations and human readable display
* Downgrade `@nextcloud/moment` to avoid compilation warning
* Set max height of NcRichContenteditable to 300px, fix top/bottom borders not visible in some cases
* Set background color of modal close button to make it more visible when it overlaps with content while scrolling

Open questions:

* In the image gen and STT pickers, should we just use the `openAssistantForm` function?


julien-nc:
This will probably get more polishing and fixes in a future PR.

PR #43, opened 2024-02-19
Add task history

julien-nc:
Add task history as a list that can be toggled at the bottom of the assistant.
The list only contains tasks of the currently selected type.
With the context menu, tasks can be deleted, canceled (when status is scheduled) and repeated (try again if they were successful).
Clicking on a task item loads it in the form (inputs + output) so it can be viewed, edited and launched again.
The task list is visible in the main assistant modal, the one when opening a task result from a notification and in the dedicated assistant page showing a task result.

image

@nimishavijay @jancborchardt I will deploy this on tech-preview soon. Any first feedback?

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -13,6 +13,8 @@ return [
 		['name' => 'assistant#scheduleTextProcessingTask', 'url' => '/task/schedule', 'verb' => 'POST'],
 		['name' => 'assistant#runOrScheduleTextProcessingTask', 'url' => '/task/run-or-schedule', 'verb' => 'POST'],
 		['name' => 'assistant#parseTextFromFile', 'url' => '/parse-file', 'verb' => 'POST'],
+		['name' => 'assistant#deleteTask', 'url' => '/task/{metaTaskId}', 'verb' => 'DELETE'],
+		['name' => 'assistant#cancelTask', 'url' => '/task/cancel/{metaTaskId}', 'verb' => 'PUT'],
 
 		['name' => 'Text2Image#processPrompt', 'url' => '/i/process_prompt', 'verb' => 'POST'],
 		['name' => 'Text2Image#getPromptHistory', 'url' => '/i/prompt_history', 'verb' => 'GET'],
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -8,6 +8,7 @@ return [
 
 		['name' => 'assistant#getAssistantTaskResultPage', 'url' => '/task/view/{metaTaskId}', 'verb' => 'GET'],
 		['name' => 'assistant#getAssistantTask', 'url' => '/task/{metaTaskId}', 'verb' => 'GET'],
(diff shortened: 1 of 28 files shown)


julien-nc:
@nimishavijay @jancborchardt This is now available on tech-preview. On this instance, most task types are handled by the OpenAI provider behind the scene.

nimishavijay:
Super nice feature! :) glad to know we are working towards a nice full fledged AI assistant :)  

**Placement and button:**
- [ ] I'm thinking rather than toggling this in a rather small modal, we could possibly replace the content of the modal itself? so the task history can take up the full modal height and width, and on top there could be a `[chevron-left] Back to Nextcloud Assistant` button
- [ ] The button itself can go on the bottom left
image
- [ ] I also see that we are using the normal size modal which is pretty small, esp. on large screens, what do you think about using a large sized modal? :) 
- [ ] in the button, "Task history" --> "Previous tasks"
- [ ] Use the [history MDI icon](https://fonts.google.com/icons?selected=Material%20Icons%20Round%3Ahistory%3A)


**Task list**
- [ ] We can make the title and subline more descriptive by including the type of task also (ref the original mockups where it looked something like this:)
image

So along the same lines, each task in the list can be of the format
```
[Task name in bold] [Input in --color-main-text]             [relative time]
[Output in --color-text-maxcontrast] 
```
- [ ] We can also think about including more actions for each task. for example, there could be a copy action on the left of the 3 dot menu that is shown on hover 

Nice-to-haves:
- [ ] Do we know the percentage completion for the in-progress tasks? Then we could use a circular progress bar as the icon (similar to this [tasks issue](https://github.com/nextcloud/tasks/pull/2464))
- [ ] Would it be possible to show the thumbnails for the images?
- [ ] The error messages seem pretty cryptic now. Is there any way to make them simpler or more actionable? In the worst case we can also just show "⚠️ Failed" and have a button which can toggle the error message

MB-Finski (approved):

julien-nc:
@nimishavijay Thanks a lot for the feedback!

>  I'm thinking rather than toggling this in a rather small modal, we could possibly replace the content of the modal itself?

But then I think it's harder to switch task types when looking at the history. The history only contains tasks of the currently select task type.
So while looking at the history, one would need to click the back button, change the task type and click the "Previous tasks" button.
What do you think?

> The button itself can go on the bottom left

The problem there is that there might be more buttons than the submit, copy and retry ones. It is possible to "integrate" the assistant in other apps so they can open it and "inject" buttons for custom actions:
image

We could have the task history button on the left and all the other one wrapping on the right. Something like:
image
What do you think?

> We can make the title and subline more descriptive by including the type of task

The task list is restricted to the current selected task type so all tasks have the same type. Would you say this needs to change and we should have all tasks in the list?

> I also see that we are using the normal size modal which is pretty small, esp. on large screens, what do you think about using a large sized modal? :)

:+1: 

> in the button, "Task history" --> "Previous tasks"

:+1: 

> Use the history MDI icon

If the button stays togglable then the chevron might be more explicit.

I'll make the rest of your suggested changes soon.

PR #45, opened 2024-03-04
enh: custom UI for scoped context chat

kyteinsky:
Design feedback required on the selector UI (Select File/Folder-Select Providers). It looks a bit too big imo. The same selector is present in speech-to-text UI so that would also need to be changed to maintain consistency. Something less obstructing/with less occupancy on the dialog is what I imagine would be a good fit.

@nimishavijay @jancborchardt 

image

### DIFF
--- a/lib/Service/AssistantService.php
+++ b/lib/Service/AssistantService.php
@@ -22,6 +22,7 @@ use OCP\IL10N;
 use OCP\Lock\LockedException;
 use OCP\PreConditionNotMetException;
 use OCP\SpeechToText\ISpeechToTextManager;
+use OCP\TextProcessing\Exception\TaskFailureException;
 use OCP\TextProcessing\FreePromptTaskType;
 use OCP\TextProcessing\IManager as ITextProcessingManager;
 use OCP\TextProcessing\ITaskType;
@@ -132,6 +133,25 @@ class AssistantService {
 					if (count($inputs) !== 2) {
 						throw new \Exception('Invalid input(s)');
 					}
+					break;
+				}
+			case 'OCA\\ContextChat\\TextProcessing\\ContextChatTaskType':
+				{
+					if ((count($inputs) !== 1 && count($inputs) !== 4)
+						|| !isset($inputs['prompt'])
+						|| !is_string($inputs['prompt'])
+					) {
+						throw new \Exception('Invalid input(s)');
+					}
+
+					if (count($inputs) === 4) {
+						if (!isset($inputs['scopeType']) || !is_string($inputs['scopeType'])
+							|| !isset($inputs['scopeList']) || !is_array($inputs['scopeList'])
+							|| !isset($inputs['scopeListMeta']) || !is_array($inputs['scopeListMeta'])) {
+							throw new \Exception('Invalid input(s)');
+						}
+					}
+
 					break;
 				}
 			default:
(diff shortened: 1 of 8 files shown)


jancborchardt (commented):
Scoped Context Chat and Context Chat don't need to be separate entries, they could be the same?

And Context Chat could just have a button on the right of the heading+subline saying e.g. "Pick folder or storage" which then opens an action menu.

@nimishavijay for more specific proposals with mockup, if you have firther ideas? :)

src/components/AssistantFormInputs.vue
```diff
@@ -256,17 +289,31 @@ export default {
 				},
 			)
 		},
+		onUpdateCc() {
```
julien-nc, line 292:
```suggestion
		onUpdateContextChat() {
```
Maybe more explicit names are needed here.

src/components/AssistantFormInputs.vue
```diff
@@ -256,17 +289,31 @@ export default {
...
+				},
+			)
+		},
 	},
 }
 
 </script>
 
-<style land="scss" scoped>
+<style lang="scss" scoped>
```
julien-nc, line 309: :+1:  :see_no_evil:

lib/Service/AssistantService.php
```diff
@@ -132,6 +133,28 @@ private function sanitizeInputs(string $type, array $inputs): array {
...
+
+					if (count($inputs) === 3) {
+						if (!isset($inputs['scopeType']) || !is_string($inputs['scopeType'])
+							|| !isset($inputs['scopeList']) || !is_array($inputs['scopeList'])) {
+							throw new \Exception('Invalid input(s)');
+						}
+						$inputs['scopeList'] = array_filter(
+							array_map(fn (array $scope) => $scope['id'] ?? null, $inputs['scopeList']),
+							fn ($scope) => is_string($scope),
+						);
```
julien-nc, line 155: Will the scope list be loaded correctly when displaying an existing task?
Wouldn't it be simpler to keep the structure of `$inputs['scopeList']` intact so it is straightforward to load it in `<ContextChatInputForm>`?

src/components/AssistantFormInputs.vue
```diff
@@ -57,6 +57,12 @@
 		</div>
 	</div>
 	<div v-else class="assistant-inputs">
+		<div v-if="selectedTaskTypeId === 'OCA\\ContextChat\\TextProcessing\\ContextChatTaskType'" class="assistant-inputs">
+			<NcCheckboxRadioSwitch :checked.sync="sccEnabled">
+				{{ t('assistant', 'Reduce context') }}
+			</NcCheckboxRadioSwitch>
+			<ContextChatInputForm v-if="sccEnabled" v-model="sccData" />
```
julien-nc, line 64:
```suggestion
			<ContextChatInputForm v-if="sccEnabled" :scc-data.sync="sccData" />
```
But maybe you had good reasons to use v-model.

julien-nc (commented):
LGTM, a few questions.

lib/Service/AssistantService.php
kyteinsky: hmm, it did work in the Assistant modal but noticed just now that it does not when pulled from the result notification. Will de-couple the UI and actual data shortly.

src/components/AssistantFormInputs.vue
kyteinsky: not really, would switch to this for consistency. Just discovered that they do the same thing.

AndyScherzinger:
Just as a side note, can't tell where mimetype icons are fetched from, we are currently in the process of updating the icons throughout the system(s) server and clients, see https://github.com/nextcloud/server/issues/44340 targeting the v29.0.0 release - fyi @kyteinsky @julien-nc

kyteinsky:
@jancborchardt new screenshots with the suggested changes:

Screenshot from 2024-03-22 15-14-53
Screenshot from 2024-03-22 15-14-40

@AndyScherzinger Thanks for the heads up! I think the only svg icon we're using is folder.svg from `/apps/theming/img/core/filetypes/folder.svg`. This seems unaffected/the latest. Others are procured from the php side using the preview manager. So we're good.

julien-nc (approved):

PR #46, opened 2024-03-06
Check all task types availability and improve app description

julien-nc:
* mention more assistant features in the readme and info.xml
* move many endpoints to OCS (anticipate integration by other apps)
* Compute the task type list on the backend side so we can check if there actually are available providers


marcelklehr (approved):

marcelklehr:
:elephant:

PR #47, opened 2024-03-08
Show empty content if no AI provider was found

julien-nc:
* Still show the top-right assistant icon if no provider is available
* If task type list is empty (so we know there is no provider available), show this in the assistant modal to admins:
image
and this to non-admins:
image

@jancborchardt Any suggestion for the title, description and admin hint?


julien-nc:
TODO:
* link to future AI doc
* for users: add "please contact your admin"

PR #49, opened 2024-03-15
Make task list and assistant form mutually exclusive

julien-nc:
Here is a first draft implementation to make the task list replace the assistant input form:

https://github.com/nextcloud/assistant/assets/11291457/35c60198-7d38-4105-bf9e-8b7838c4b641

For now the task list is still filtered to only show tasks which type is the selected one.
@jancborchardt @nimishavijay Do you really think it's better to also get rid of the task type selector and show all tasks?
Any other feedback on how it looks?


julien-nc:
We can change or polish this later.

PR #52, opened 2024-03-27
Add OpenAPI specs

julien-nc:
* Change Php namespace from TpAssistant to Assistant
* Endpoints that should not be used by clients: move to non-ocs controller
* Change the frontend namespace (keep the old one to avoid breaking existing integrations)
* Add OpenAPI specs

### DIFF
--- a/lib/Controller/Text2ImageApiController.php
+++ b/lib/Controller/Text2ImageApiController.php
@@ -3,26 +3,31 @@
 // SPDX-FileCopyrightText: Sami Finnilä <sami.finnila@nextcloud.com>
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-namespace OCA\TpAssistant\Controller;
+namespace OCA\Assistant\Controller;
 
 use Exception;
-use OCA\TpAssistant\Service\Text2Image\Text2ImageHelperService;
+use OCA\Assistant\ResponseDefinitions;
+use OCA\Assistant\Service\Text2Image\Text2ImageHelperService;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\Attribute\AnonRateLimit;
 use OCP\AppFramework\Http\Attribute\BruteForceProtection;
 use OCP\AppFramework\Http\Attribute\NoAdminRequired;
 use OCP\AppFramework\Http\Attribute\NoCSRFRequired;
+use OCP\AppFramework\Http\Attribute\OpenAPI;
 use OCP\AppFramework\Http\Attribute\PublicPage;
 use OCP\AppFramework\Http\DataDisplayResponse;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCSController;
-use OCP\Db\Exception as DbException;
 
 use OCP\Files\NotPermittedException;
 use OCP\IL10N;
 use OCP\IRequest;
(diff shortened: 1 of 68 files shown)


lib/Controller/Text2ImageApiController.php
```diff
@@ -119,21 +114,31 @@ public function getImage(string $imageGenId, int $fileNameId): DataDisplayRespon
...
 	#[BruteForceProtection(action: 'imageGenId')]
+	#[OpenAPI(scope: OpenAPI::SCOPE_DEFAULT, tags: ['image_generation'])]
 	public function getGenerationInfo(string $imageGenId): DataResponse {
 		try {
 			$result = $this->text2ImageHelperService->getGenerationInfo($imageGenId, $this->userId, true);
 		} catch (Exception $e) {
-			$response = new DataResponse(['error' => $e->getMessage()], (int) $e->getCode());
-			if ($e->getCode() === Http::STATUS_BAD_REQUEST || $e->getCode() === Http::STATUS_UNAUTHORIZED) {
+			/** @var Http::STATUS_BAD_REQUEST|Http::STATUS_NOT_FOUND|Http::STATUS_INTERNAL_SERVER_ERROR $exceptionCode */
+			$exceptionCode = (int) $e->getCode();
```
marcelklehr, line 137: Wondering if there is a case where $exceptionCode is not a valid http status code...

lib/Controller/Text2ImageController.php
```diff
@@ -43,4 +54,23 @@ public function showGenerationPage(?string $imageGenId, ?bool $forceEditMode = f
 
 		return new TemplateResponse(Application::APP_ID, 'imageGenerationPage');
 	}
+
+	/**
+	 * @return DataResponse
+	 */
+	#[NoAdminRequired]
+	#[NoCSRFRequired]
+	public function getPromptHistory(): DataResponse {
```
marcelklehr, line 63: History is not part of the API?

src/main.js
```diff
@@ -7,20 +7,23 @@ import { subscribe } from '@nextcloud/event-bus'
...
  * - Listen to notification event
  */
 function init() {
-	if (!OCA.TPAssistant) {
+	if (!OCA.Assistant) {
 		/**
 		 * @namespace
 		 */
-		OCA.TPAssistant = {
+		OCA.Assistant = {
```
marcelklehr, line 19: Are there docs anywhere that need to be changed as well?

marcelklehr (approved):

lib/Controller/Text2ImageApiController.php
julien-nc: As we are the ones throwing the exception, we know which codes we can get here. I don't like how it's implemented, I just made the necessary changes for the psalm check to be green. This would definitely need a bit of refactoring/cleaning.

lib/Controller/Text2ImageController.php
julien-nc: Not the prompt history, it is only used by the smart pickers.
This is not related with task history.

src/main.js
julien-nc: Working on it :grin:

lib/Controller/Text2ImageController.php
marcelklehr: ah

PR #58, opened 2024-04-08
fix(ContextChat): update inputs on selective context checkbox change

kyteinsky:


kyteinsky:
@julien-nc Not sure if this is the best approach but it fixes the issue

julien-nc:
If you wanna do it in a similar fashion than the other task types, the change/checked/update events on the sccEnabled checkbox and `<ContextChatInputForm>` could call `onUpdateContextChat`.

Also, I think `onUpdateContextChat` should be called in the `selectedTaskTypeId` watch method when context chat is selected.

julien-nc (approved):

PR #60, opened 2024-04-10
chore: add node and eslint gh workflows

kyteinsky:


julien-nc:
Why the node action? We don't include compiled assets in this repo.

kyteinsky:
To check if the build passes, we discard the compiled js then. And lint checks.

julien-nc:
Ok why not :grin:

julien-nc (approved):

PR #63, opened 2024-04-12
Add empty content when history is empty

### LINKED ISSUE #62: Past tasks button should be hidden if there are no old tasks

c.nc.c

c.nc.c

Firefox

### Describe the Bug

The button to see past tasks is always shown, even when there are no tasks to show.

### Expected Behavior

The button should only be shown if there are actually tasks that would be displayed there.

### To Reproduce

1. Fresh Nextcloud instance (with some task providers)
2. Open Assistant
3. Select any of the tasks and the button will be shown

julien-nc:
closes #62
The history task list is retrieved after the history button is clicked so we can't hide the button if the history is empty.
But we can show an empty content:

image

### DIFF
--- a/src/components/AssistantTextProcessingForm.vue
+++ b/src/components/AssistantTextProcessingForm.vue
@@ -32,7 +32,7 @@
 			</div>
 			<TaskList
 				class="history--list"
-				:task-type="mySelectedTaskTypeId"
+				:task-type="selectedTaskType"
 				:loading.sync="historyLoading"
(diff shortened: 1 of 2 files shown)


provokateurin (approved):
LGTM, but maybe the issue should be kept open so that it can be implemented in the future.

julien-nc:
Since it's possible to change the task type while seeing the history, maybe it's still good to see the history button even if it leads to an empty history.

It might seem weird to users not to find the history button.

PR #66, opened 2024-04-15
Add rtf support 

julien-nc:
None of the Php RTF parsers I tried work.
Thankfully `henck/rtf-to-html` works fine.
So this uses `henck/rtf-to-html` to convert to html and then `html2text/html2text` to convert to text.


kyteinsky (approved):
works perfectly! Thanks for the quick resolution.

PR #72, opened 2024-04-17
Add standalone assistant page

julien-nc:
This makes the task result page able to show an empty task. It is reachable at `apps/assistant/`. This might be useful for clients that want to open the "raw" assistant in a browser.


kyteinsky (approved):
both routes work nicely :tada:

PR #94, opened 2024-06-27
button disable fix + make fallback strings same for translation ease

kyteinsky:
+ edit action buttons alignment
+ security fixes in chatty ui controller

### DIFF
--- a/src/components/ChattyLLM/Message.vue
+++ b/src/components/ChattyLLM/Message.vue
@@ -90,7 +90,7 @@ export default {
 	data: () => {
 		return {
 			displayName: getCurrentUser()?.displayName ?? getCurrentUser()?.uid ?? t('assistant', 'You'),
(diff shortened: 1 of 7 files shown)


src/components/ChattyLLM/Message.vue
```diff
@@ -90,7 +90,7 @@ export default {
 	data: () => {
 		return {
 			displayName: getCurrentUser()?.displayName ?? getCurrentUser()?.uid ?? t('assistant', 'You'),
-			userId: getCurrentUser()?.uid ?? t('assistant', 'yooniquely-you'),
+			userId: getCurrentUser()?.uid ?? t('assistant', 'You'),
```
julien-nc, line 93: If there is no user in the context, something is wrong, it does not matter that much to have a fallback there, right? Especially since this userId is only passed to the avatar. Anyway, can't hurt :+1:

julien-nc (approved):
:+1:

PR #108, opened 2024-08-12
chore: More links to docs

marcelklehr:


src/components/AdminSettings.vue
```diff
@@ -59,8 +68,12 @@
 						<div v-if="!state.speech_to_text_picker_available" class="checkbox-text">
 							<InformationOutlineIcon class="icon" />
 							<span>
-								{{ t('assistant', 'To enable this feature, please install a speech-to-text provider.') }}
+								{{ t('assistant', 'To enable this feature, please install a speech-to-text provider:') }}
```
julien-nc, line 71:
```suggestion
								{{ t('assistant', 'To enable this feature, please install a speech-to-text provider') + ':' }}
```

julien-nc (changes requested):
Small detail and good to go. Thanks!

src/components/AdminSettings.vue
marcelklehr: huh. Not sure if this will have the intended effect in all languages (ie. RTL?)

PR #124, opened 2024-09-16
fix: render context chat's the referenced source items

kyteinsky:


kyteinsky:
Not sure if there is a cleaner way. Should we move the entire output render to context chat's form?

src/components/AssistantTextProcessingForm.vue
```diff
@@ -47,12 +47,20 @@
...
 					:shape-options="selectedTaskType.outputShapeEnumValues ?? null"
 					:optional-shape-options="selectedTaskType.optionalOutputShapeEnumValues ?? null"
 					:values.sync="myOutputs"
 					:show-advanced.sync="showAdvanced" />
+				<div v-if="mySelectedTaskTypeId === 'context_chat:context_chat'" style="margin-top: 8px">
+					{{ t('assistant', 'The sources referenced to generate the above response') }}
+					<ContextChatOutputSources
+						v-if="mySelectedTaskTypeId === 'context_chat:context_chat'"
+						:sources="myOutputs.sources" />
+				</div>
```
julien-nc, line 63: Let's put as few Context-Chat-specific stuff as possible in this file. The label text can go in the `ContextChatOutputSources` component.

Also we can do something similar than for the input form. In this file, the output `TaskTypeFields` component and this new `div` could be moved in a `AssistantFormOutputs` component.

Wdyt?

julien-nc (changes requested):

julien-nc (approved):

PR #135, opened 2024-09-30
Get rid of a few warnings during js packing

julien-nc:
image
and
image


kyteinsky:
Nice, that's almost all of them!

```
assistant/src/components/ChattyLLM/ChattyLLMInputForm.vue
  455:9  warning  Caution: `axios` also has a named export `isCancel`. Check if you meant to write `import {isCancel} from '@nextcloud/axios'` instead  import/no-named-as-default-member
```
This can be fixed here too.

kyteinsky (approved):

marcelklehr (approved):

PR #137, opened 2024-10-02
enh: Report task ID in failure message

marcelklehr:


julien-nc (changes requested):
Noice!
This could also be done in the `openAssistantTask` function of `assistant.js` and in `AssistantPage.vue` where we also react to failing tasks.

Issue #138, opened 2024-10-02
Missing worker policy when launching assistant


smarinier:
### Which version of assistant are you using?

2.1.0

### Which version of Nextcloud are you using?

30.0.0

### Which browser are you using? In case you are using the phone App, specify the Android or iOS version and device please.

Chrome  128.0.6613.114 

### Describe the Bug

Launch the assistant:

ScheduledEmptyConten…KXdH-J.chunk.mjs:33 Uncaught (in promise) SecurityError: Failed to construct 'Worker': Access to the script at 'blob:

Note: This looks like a cross application policy trouble (hidden by a classical use of other application). See https://github.com/nextcloud/server/issues/39849

### Expected Behavior

Launch the assistant

### To Reproduce

A NextCloud 30 without any application modifying the policy in a convenient way for Assistant (eg Talk). (A minimal Nextcloud)

PR #139, opened 2024-10-02
fix: #138 Missing worker policy when launching assistant

smarinier:
See also https://github.com/nextcloud/server/issues/39849


julien-nc:
Hi,
Thanks a lot for the PR.
I can reproduce on Chromium but everything works fine on Firefox, there's only this console error message: `Content-Security-Policy: The page’s settings blocked a worker script (worker-src) at blob:https://nextcloud.local/a9f6a3b5-2522-4efe-939a-28a3a33ba63f from being executed because it violates the following directive: “worker-src 'self'”`

julien-nc (approved):

PR #153, opened 2024-11-14
Change 'Generate text' to 'Ask anything'

julien-nc:
@jancborchardt Or maybe "Generate text" is already good enough but it needs to be adopted in the clients. Wdyt?

### DIFF
--- a/lib/Service/AssistantService.php
+++ b/lib/Service/AssistantService.php
@@ -181,7 +181,7 @@ class AssistantService {
 			$taskTypeArray['priority'] = self::TASK_TYPE_PRIORITIES[$typeId] ?? 1000;
 
 			if ($typeId === TextToText::ID) {
-				$taskTypeArray['name'] = $this->l10n->t('Generate text');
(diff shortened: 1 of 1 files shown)


jancborchardt:
@Hyeyoung346 which wording would you prefer?

And yes for sure, whichever we choose needs to be the same among web and mobile. @tobiasKaminsky

Hyeyoung346:
I prefer “Ask anything” because we could attach photos, files, voice  in the input process and output also could be anything if we could expand more functionality in the future. Currently only text input/output is available, but I think “Generate Text” limits future possibilities.  @julien-nc @jancborchardt

tobiasKaminsky:
> And yes for sure, whichever we choose needs to be the same among web and mobile. @tobiasKaminsky

This is exposed via API, so clients will be okay.

jancborchardt (approved):

janepie:
Closing as the "ask anything" version is "Chat with AI" nowadays, I would say

PR #156, opened 2024-11-20
Fix polling new chat message when switching sessions

julien-nc:
closes #154
(This replaces it)

Add a new endpoint to check if a session is in a "thinking" state (meaning it currently has a scheduled/running task).
Use this endpoint to know if we should start polling when switching sessions.

Extra changes:

* Prevent scheduling multiple llm tasks for one session
* Fix using incorrect session ID in run/poll methods in the frontend

### DIFF
--- a/src/components/ChattyLLM/ChattyLLMInputForm.vue
+++ b/src/components/ChattyLLM/ChattyLLMInputForm.vue
@@ -219,6 +219,7 @@ export default {
 
 	watch: {
 		async active() {
+			this.loading.llmGeneration = false
 			this.allMessagesLoaded = false
 			this.chatContent = ''
 			this.msgCursor = 0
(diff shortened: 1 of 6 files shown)


src/components/ChattyLLM/ChattyLLMInputForm.vue
```diff
@@ -219,20 +219,69 @@ export default {
 
 	watch: {
 		async active() {
+			this.loading.llmGeneration = false
+			this.loading.titleGeneration = false
```
kyteinsky, line 223: let's set them to true. For llm generation when we do a network request to check if there is a running task, it switches from false to true quickly resulting in a glitchy effect. Since we're going to do this check on every session change, let's first assume there is a task and block the input. Later we free the input when there are no tasks.

src/components/ChattyLLM/ChattyLLMInputForm.vue
```diff
@@ -219,20 +219,69 @@ export default {
...
+						console.error('onCheckSessionTitle error:', error)
+						showError(error?.response?.data?.error ?? t('assistant', 'Error getting the generated title for the conversation'))
+					} finally {
+						this.loading.titleGeneration = false
+					}
+				}
+			} catch (error) {
+				console.error('check session error:', error)
+				showError(t('assistant', 'Error checking if the session is thinking'))
 			}
```
kyteinsky, line 285: here in a `finally` block

src/components/ChattyLLM/ChattyLLMInputForm.vue
```diff
@@ -555,46 +604,64 @@ export default {
...
-						{ params: { taskId, sessionId: this.active.id } },
+						{ params: { taskId, sessionId } },
 					).then(response => {
 						clearInterval(this.pollMessageGenerationTimerId)
-						resolve(response.data)
+						if (sessionId === this.active.id) {
+							resolve(response.data)
+						} else {
+							console.debug('Ignoring received message for session ' + sessionId + ' that is not selected anymore')
+							// should we reject here?
```
kyteinsky, line 619: let's clear the interval at line 610 when the session is not the same. Polling starts on session switch again anyway.

src/components/ChattyLLM/ChattyLLMInputForm.vue
```diff
@@ -555,46 +604,64 @@ export default {
...
+							console.debug('Ignoring received title for session ' + sessionId + ' that is not selected anymore')
+							// should we reject here?
+						}
 						clearInterval(this.pollTitleGenerationTimerId)
-						resolve(response)
 					}).catch(error => {
+						if (sessionId !== this.active.id) {
+							console.debug('Stop polling title for session ' + sessionId + ' because it is not selected anymore')
+							clearInterval(this.pollTitleGenerationTimerId)
+						}
```
kyteinsky, line 657: same here

src/components/ChattyLLM/ChattyLLMInputForm.vue
```diff
@@ -219,20 +219,69 @@ export default {
...
+						console.debug('checkTaskPolling result:', message)
+						this.messages.push(message)
+						this.scrollToBottom()
+					} catch (error) {
+						console.error('checkGenerationTask error:', error)
+						showError(t('assistant', 'Error generating a response'))
+					} finally {
+						this.loading.llmGeneration = false
+					}
+				} else if (checkSessionResponse.data.titleTaskId !== null) {
```
kyteinsky, line 262: `if` would be good here since there might be message and title tasks both scheduled

kyteinsky (approved):

PR #167, opened 2024-12-18
feat: save default target language & set detect_language as origin default

janepie:
This adds a short loading time before task types will be shown in the assistant modal. Ideas how to prevent that very welcome :)

### DIFF
--- a/src/components/AssistantTextProcessingForm.vue
+++ b/src/components/AssistantTextProcessingForm.vue
@@ -347,19 +347,6 @@ export default {
 				fileId,
 			})
 		},
-
-		async getLastTargetLanguage() {
-			const { default: axios } = await import('@nextcloud/axios')
-			const { generateUrl } = await import('@nextcloud/router')
-
-			const req = {
-				params: {
-					key: 'last_target_language',
-				},
-			}
-			const url = generateUrl('/apps/assistant/config')
-			return axios.get(url, req)
-		},
 		getTaskTypes() {
 			this.loadingTaskTypes = true
 			axios.get(generateOcsUrl('/apps/assistant/api/v1/task-types'))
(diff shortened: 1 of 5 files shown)


src/components/AssistantTextProcessingForm.vue
```diff
@@ -365,27 +378,38 @@ export default {
...
-							tt.inputShape.input.placeholder = t('assistant', 'landscape trees forest peaceful')
-							tt.inputShape.numberOfImages.placeholder = t('assistant', 'a number')
-						} else if (tt.id === 'core:contextwrite' && tt.inputShape.source_input && tt.inputShape.style_input) {
-							tt.inputShape.style_input.placeholder = t('assistant', 'Shakespeare or an example of the style')
-							tt.inputShape.source_input.placeholder = t('assistant', 'A description of what you need or some original content')
-						}
-					})
-					this.taskTypes = taskTypes
+
+					this.getLastTargetLanguage()
```
julien-nc, line 382: It would probably work to call this asynchronously and change the task type (in `this.taskTypes`) once you get the answer.

It would mean keeping the previous code (`taskTypes.forEach` and then `this.taskTypes = taskTypes`) and then doing the `this.getLastTargetLanguage().then(r => // find and change the task type in this.taskTypes)`

This would solve the delay issue as the `taskTypes` data prop would be set as soon as possible. It would not be slowed down by `this.getLastTargetLanguage()`.

julien-nc (changes requested):
Recommendation for the delay issue.

julien-nc (approved):
Added a small fix and switched to using initial state + an OCA.Assistant variable to handle the last target language.

marcelklehr (approved):

PR #170, opened 2024-12-20
Implement pdf content extraction

dhx:
Currently pdf files can't be used as a text source in the assistant (while they work in context chat).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
 		"erusev/parsedown": "^1.7",
 		"henck/rtf-to-html": "^1.2",
 		"html2text/html2text": "^4.3",
-		"phpoffice/phpword": "^1.2"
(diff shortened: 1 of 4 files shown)


dhx:
Hi @julien-nc I see you've added rtf support, so maybe you can take a look if this implementation for pdf looks good to you or if something is missing?

julien-nc (changes requested):
Looks good!

Can you rebase on the main branch?
There's a conflict with composer.lock. Just get rid of yours, rebase on main, install `smalot/pdfparser` and make a new commit.

julien-nc:
I made another PR with some adjustments. See #204
Thanks for your PR though!

dhx:
@julien-nc thanks for taking it over!

PR #180, opened 2025-01-23
Fix: delete assistant modal title

### LINKED ISSUE #118: UI interferes with main Nextcloud navigation

Daily testing server on 2024-08-21 (version not exposed to non-admin user)

31.0.0 dev

Firefox 129.0 (Linux)

### Describe the Bug

The title of the Assistant interface overlays and interferes with Nextcloud's main navigation UI

image

image

### Expected Behavior

Expecting the Nextcloud navigation UI to be always available and visually unobstructed.

### To Reproduce

* open Assistant
* ...voila!

janepie:
fixes #118 

keeps the string as labelId for accessibility reasons

### DIFF
--- a/src/components/AssistantTextProcessingModal.vue
+++ b/src/components/AssistantTextProcessingModal.vue
@@ -7,6 +7,7 @@
 		:size="modalSize"
 		:can-close="false"
 		:name="t('assistant', 'Nextcloud Assistant')"
+		dark
 		class="assistant-modal"
 		@close="onCancel">
 		<div ref="modal_content"


julien-nc (approved):
This is the same for all full-size modals displayed in NC. If we decide it does not make sense to display a title for those, it should probably be changed in `@nextcloud/vue`.

Anyway, this fix is fine.

src/components/AssistantTextProcessingModal.vue
```diff
@@ -6,7 +6,7 @@
 	<NcModal v-if="show"
 		:size="modalSize"
 		:can-close="false"
-		:name="t('assistant', 'Nextcloud Assistant')"
+		:label-id="t('assistant', 'Nextcloud Assistant')"
```
marcelklehr, line 9: I'm not sure if labelId is the right choice for this, as it takes an element id, I think, not a free-form string.

janepie: Ah damn you're right. i just checked, I cannot give the modal an aria label directly, it has to be through `name` or another element. What's the best way to work with that, creating a dummy element somehow?

julien-nc: You can try setting an `aria-label` to the modal component and check it appears in the UI.

marcelklehr: git push? :)

janepie: uuuhm sorry, done 😅 but we might also wait for https://github.com/nextcloud-libraries/nextcloud-vue/pull/6434 with this, wdyt?

julien-nc: The modal opacity change does not require any change in the assistant so we can merge this PR IMO.

janepie: No, but with that we could keep the title (do we want to keep it?)

julien-nc: Oh right, if the frame is not transparent anymore, then the issue is gone. Let's wait and close this PR if https://github.com/nextcloud-libraries/nextcloud-vue/pull/6434 goes in.

It would be weird to remove the title while the frame is transparent and add it again when the frame becomes opaque.

src/components/AssistantTextProcessingModal.vue
```diff
@@ -6,7 +6,7 @@
 	<NcModal v-if="show"
 		:size="modalSize"
 		:can-close="false"
-		:name="t('assistant', 'Nextcloud Assistant')"
+		:aria-label="t('assistant', 'Nextcloud Assistant')"
```
susnux, line 9: You do not have to wait for the fix you can already force it:

```suggestion
		dark
		:name="t('assistant', 'Nextcloud Assistant')"
```

janepie: Good to know, thanks!

julien-nc (approved):
:ok_hand:

PR #186, opened 2025-01-27
Fix choose file button position with RTL languages

julien-nc:
### LTR

image

### RTL Before

image

### RTL After

image

For some reason left must be unset when right is defined and the other way around.

cc @DaphneMuller


janepie (approved):
Looks good!

PR #188, opened 2025-02-03
Fix regeneration of freshly generated chat messages

julien-nc:
### Short version

This fixes regeneration of messages that were freshly generated. The task ID is now stored in the message entry so we can get a message from a task ID. That is helpful for the `/chat/check_generation?taskId=123` endpoint.

### Long version

When a chat task has finished, a listener stores a message in the `oc_assistant_chat_msgs` table.
In the UI, when a message is currently being generated, the `/chat/check_generation?taskId=123` assistant endpoint is polled.
When the related controller sees that the task has finished and is successful:
* it used to generate and return a fake message object on the fly, based on the data in the finished task (because it could not get the message in the DB based on the task ID)
* it now gets the real message that was stored by the listener

The problem was that the message obtained via `/chat/check_generation?taskId=123` had no ID so it could not be regenerated because we do it with a request like `/chat/regenerate?messageId=619&sessionId=48` but we didn't have an ID for the freshly generated messages.

The issue was not so visible because when opening the assistant, the message list was correctly loaded (from DB entries), with all message having an ID.


lib/Migration/Version020400Date20250203125359.php
```diff
@@ -0,0 +1,44 @@
...
+		if ($schema->hasTable('assistant_chat_msgs')) {
+			$table = $schema->getTable('assistant_chat_msgs');
+			if (!$table->hasColumn('task_id')) {
+				$table->addColumn('task_id', Types::BIGINT, [
+					'notnull' => true,
+					'default' => 0,
+					'unsigned' => true,
+				]);
+				$schemaChanged = true;
+			}
```
kyteinsky, line 39: wdyt about making it unique and adding an index here for just the task_id?

lib/Db/ChattyLLM/MessageMapper.php
```diff
@@ -101,6 +101,24 @@ public function getMessageById(int $messageId): Message {
...
+	 * @throws \OCP\DB\Exception
+	 */
+	public function getMessageByTaskId(int $sessionId, int $taskId): Message {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select(Message::$columns)
+			->from($this->getTableName())
+			->where($qb->expr()->eq('session_id', $qb->createPositionalParameter($sessionId, IQueryBuilder::PARAM_INT)))
+			->andWhere($qb->expr()->eq('task_id', $qb->createPositionalParameter($taskId, IQueryBuilder::PARAM_INT)));
+
+		return $this->findEntity($qb);
```
kyteinsky, line 119: only task_id should suffice here and the index should help
maybe it can be argued that regeneration would not be called that often so it is also fine to skip it.

julien-nc: If we do this and don't add extra checks in `ChattyLLMController` we allow any user to get any message as long as they provide a valid session ID. They can ask for any task id.

So either we keep it like i did, or we do as you say but check that the message object we get has the correct session ID.

lib/Migration/Version020400Date20250203125359.php
julien-nc: :+1: I'll add an index. I don't know now which one, depending on the decision we make on your other comment.

lib/Db/ChattyLLM/MessageMapper.php
kyteinsky: yup, you're right, let's keep sessionId

lib/Migration/Version020400Date20250203125359.php
julien-nc: index added

kyteinsky (approved):
:100:

PR #194, opened 2025-02-13
Adjustments of agency confirmation style

julien-nc:
### Before

image

### After

image

### After 2

image

### After 3
image

For actions with a lot of params:

image

image

More will come.


marcelklehr:
Looks nice!

jancborchardt (approved):
Wow @julien-nc, that looks super good already! :) If it is different options, "After 3" is the best.

Small point I don’t know if I mentioned in our call before: Usually we have the primary button on the right, and cancel on the left (or left of it). Could the buttons be switched around?

julien-nc:
@jancborchardt Thanks for the feedback! Buttons have been switched. New release incoming.

PR #218, opened 2025-04-09
Better naming of shared output files, support more mime types

julien-nc:
When a task output file is shared from the assistant UI, the file extension is guessed from the mimetype. More types are now supported.

For the download endpoint, setting the mimetype correctly in the response headers is enough for browsers to guess the extension in the suggested file name.

### DIFF
--- a/lib/Service/AssistantService.php
+++ b/lib/Service/AssistantService.php
@@ -439,27 +439,16 @@ class AssistantService {
 	 * @throws NotPermittedException
 	 */
 	private function getTargetFileName(File $file): string {
-		$mime = mime_content_type($file->fopen('rb'));
-		$name = $file->getName();
+		$mimeType = mime_content_type($file->fopen('rb'));
+		$fileName = $file->getName();
 
-		$mime2ext = [
-			'image/png' => '.png',
-			'image/jpeg' => '.jpg',
-			'image/webp' => '.webp',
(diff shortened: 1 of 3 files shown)


lib/Service/AssistantService.php
```diff
@@ -441,12 +441,20 @@ public function shareOutputFile(string $userId, int $ocpTaskId, int $fileId): st
...
+			'image/webp' => '.webp',
+			'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet' => '.xlsx',
+			'application/vnd.oasis.opendocument.spreadsheet' => '.ods',
+			'application/vnd.openxmlformats-officedocument.wordprocessingml.document' => '.docx',
+			'application/vnd.oasis.opendocument.text-web' => '.odt',
+			'application/vnd.openxmlformats-officedocument.presentationml.presentation' => '.pptx',
+			'application/vnd.oasis.opendocument.presentation' => '.odp',
+			'application/vnd.oasis.opendocument.graphics' => '.odg',
+		];
+		$ext = $mime2ext[$mime] ?? '';
```
marcelklehr, line 457: Could be nice to use something like this https://github.com/ralouphie/mimey

marcelklehr (approved):

marcelklehr (approved):

PR #219, opened 2025-04-09
Use psalm 6.4.0 

julien-nc:
* Use psalm 6.4.0 (higher is too restrictive on Php versions and it's hard to run psalm locally)
* Add a matrix in the psalm action
* Update composer dependencies

### DIFF
--- a/.github/workflows/psalm.yml
+++ b/.github/workflows/psalm.yml
@@ -6,7 +6,7 @@
 # SPDX-FileCopyrightText: 2022-2024 Nextcloud GmbH and Nextcloud contributors
 # SPDX-License-Identifier: MIT
 
-name: Psalm static analysis
+name: Psalm Static analysis
 
 on:
   pull_request:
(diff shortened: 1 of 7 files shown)


.github/workflows/psalm.yml
```diff
@@ -37,24 +37,51 @@ concurrency:
 jobs:
   static-analysis:
     runs-on: ubuntu-latest
+    strategy:
+      # do not stop on another job's failure
+      fail-fast: false
+      matrix:
+        ocp-version: ['30', '31', 'dev-master']
+        php-version: ['8.1', '8.2', '8.3', '8.4']
```
marcelklehr, line 45: Mh, latest workflow from nextcloud/.github should get this automatically from the supported nextcloud versions, I believe

marcelklehr:
Did you have no issues with the Override attribute missing and code being unreachable?

.github/workflows/psalm.yml
julien-nc: It's now based on the nextcloud/.github psalm-matrix one. I just changed the "on" condition to only run on PRs for some branches and if some files where changed.

marcelklehr (approved):

PR #220, opened 2025-04-16
feat: display expected task runtime when running

### LINKED ISSUE #212: Show rough estimated time

The providers are already declaring a time estimation for the tasks they will handle. This is not very precise but already gives us a rough idea. It would be nice to show that estimation to the user.

In the loading view, and maybe some other places, we could show a simple approximation of the time. We can determine a threshold to split between "short" and "long" tasks, like 60 seconds. A short task's estimation would be "in a few seconds" and the long one would be "in a few minutes".

This is not fully defined yet so other ideas can be discussed 😁

edward-ly:
Resolves #212. Formatting and text may still need some adjusting, though.

20250415_17h00m05s_scrot

### DIFF
--- a/src/assistant.js
+++ b/src/assistant.js
@@ -85,6 +85,7 @@ export async function openAssistantForm({
 			view.loading = true
 			view.showSyncTaskRunning = true
 			view.progress = null
+			view.expectedRuntime = null
 			view.inputs = inputs
 			view.selectedTaskTypeId = taskTypeId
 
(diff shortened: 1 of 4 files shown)


julien-nc (commented):
The expected runtime is currently only displayed when the task is running (when the progress bar is being displayed). We get the task expected runtime value as soon as we schedule the task, before it is running. We could display it as soon as we have it. The delay between scheduling and running should not be long, i think it's fine to show the expected runtime even if the task is not running yet.

@jancborchardt @nimishavijay @edward-ly Wdyt?

julien-nc:
Hey @edward-ly . I think we can go with my suggested change. Show the expected runtime as soon as the task is scheduled (and not wait for it to be running).

edward-ly:
Done. I also removed the generic message when the runtime is not retrieved yet as it only flashed for a brief moment before the message changed afterwards.

janepie (approved):
Looks good!

jancborchardt (approved):
Nice improvement! :)

PR #221, opened 2025-04-16
Support markdown links preview in chat UI

julien-nc:
Override NcRichText link resolving to support markdown links.

This currently produces a lot of warnings because the `references` prop of `<NcRichText>` is declared as Object but the implementation actually expects an array. I'll fix that in nc/vue.

### DIFF
--- a/src/components/ChattyLLM/Message.vue
+++ b/src/components/ChattyLLM/Message.vue
@@ -37,6 +37,7 @@
 			:text="message.content || t('assistant', '(empty message)')"
 			:use-markdown="true"
 			:reference-limit="1"
+			:references="references"
 			:autolink="true" />
 	</div>
 </template>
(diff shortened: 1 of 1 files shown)


julien-nc:
Let's wait to merge that, I realized it was not avoiding the closing parenthesis. I think we need 2 regexps, one for plain URLs and one for markdown links. I'll do that later.

janepie:
Or we need to strip stuff from the end? Like if the link is part of a sentence in brackets (like this https://test.com) or at the end of a sentence with a dot e.g. https://test.com.

julien-nc:
The problem is that brackets are allowed in GET params. So why should we strip anything from (like that https://text.com?plop=lala) .
We can make the choice to not support trailing brackets (that should probably be url-encoded anyway. But how do we choose the list of characters that we strip? Most of them will be allowed in GET params. Not sure how to proceed here.

src/components/ChattyLLM/Message.vue
```diff
@@ -97,14 +103,41 @@ export default {
...
 	methods: {
 		copyMessage(message) {
 			navigator.clipboard.writeText(message)
 			showSuccess(t('assistant', 'Message copied to clipboard'))
 		},
+		fetch() {
+			const urlMatch = (new RegExp(PLAIN_URL_PATTERN).exec(this.message.content.trim()))
+			const mdMatch = (new RegExp(MARKDOWN_LINK_PATTERN).exec(this.message.content.trim()))
+			const firstMatch = urlMatch
+				? urlMatch[1].replaceAll(/[).,:;]+$/g, '')
```
janepie, line 127:
```suggestion
				? urlMatch[1].replaceAll(/[).,:;!?]+$/g, '')
```

I would add ! and ? for URLs and sentence ends

janepie (changes requested):

src/components/ChattyLLM/Message.vue
julien-nc: Done

janepie (approved):
:rocket:

PR #225, opened 2025-04-23
fix(NumberField): use number type for NcInputField

### LINKED ISSUE #215: Improve number input
Instead of a NcInputField we could use a custom component to input numbers. As they are mostly used to set the desired number of generated elements, we could have 5 buttons (1, 2, 3, 4, 5) to quickly select those values and another button labeled "custom" to enter any value.

Maybe this justifies the creation of a new TaskProcessing input shape type.

To be discussed.

edward-ly:
Resolves #215. This enables the use of up/down arrow keys to increment/decrement values in a number field, which I believe is preferable over tabbing between multiple buttons with constant values. Although, there are no clickable up/down arrow buttons visible in the input field which would help indicate to the user that the field is a number field (but that may be an issue with the component itself, I'm not sure).

The only other difference I could tell between the number and text fields is the emitted values of NaN and '', respectively, when the field is empty. Thus, we also need to prevent the form from submitting if any of the input fields is NaN (which, correct me if I'm wrong, I don't think is an allowed value in any context anyway).

### DIFF
--- a/src/components/AssistantTextProcessingForm.vue
+++ b/src/components/AssistantTextProcessingForm.vue
@@ -265,7 +265,7 @@ export default {
 						SHAPE_TYPE_NAMES.Image,
 						SHAPE_TYPE_NAMES.Audio,
 						SHAPE_TYPE_NAMES.Video,
-					].includes(fieldType) && typeof value === 'number')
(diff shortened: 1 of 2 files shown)


julien-nc:
I think the goal is to make it easier and quicker for non-tech users to set a number value. Let's discuss that further in the related issue.

PR #226, opened 2025-04-24
feat(ChattyLLM): replace loading spinner with skeleton message

### LINKED ISSUE #214: [ChatUI] Improve message generation loading indicator
For now, there is only a spinner in the submit button of the chat UI input form when generating a new message. There could be a ghost message in the message list (like when switching conversations in Talk).

edward-ly:
Resolves #214. The [LoadingPlaceholder component](https://github.com/nextcloud/spreed/blob/main/src/components/UIShared/LoadingPlaceholder.vue) from the Talk app was copied and minimally adapted to fit the chat UI. Also, the message input field already displays `Processing...` while the message is being generated, so I don't think there is a need for additional text underneath the placeholder either.

### DIFF
--- a/src/components/ChattyLLM/ConversationBox.vue
+++ b/src/components/ChattyLLM/ConversationBox.vue
@@ -18,18 +18,20 @@
 				<NcLoadingIcon />
 			</template>
 		</NoSession>
-		<Message v-for="(message, idx) in messages"
-			v-else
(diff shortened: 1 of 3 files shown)


julien-nc (approved):
Works just fine. Good that you removed the loading spinner in the submit button, it was a bit too much now with the placeholder message :+1:

PR #230, opened 2025-04-30
Feat: Information Source info in Chat UI

janepie:
The info button shows where the information the Agent used comes from
Defined per tool, every source will only be listed once

image

Needs https://github.com/nextcloud/context_agent/pull/22 to show sources

### DIFF
--- a/lib/Listener/ChattyLLMTaskListener.php
+++ b/lib/Listener/ChattyLLMTaskListener.php
@@ -61,7 +61,8 @@ class ChattyLLMTaskListener implements IEventListener {
 			$message->setRole('assistant');
 			$message->setContent(trim($task->getOutput()['output'] ?? ''));
 			$message->setTimestamp(time());
-			$message->setSources(json_encode($task->getOutput()['sources'] ?? []));
+			$sources = json_encode($task->getOutput()['sources'] ?? []);
+			$message->setSources($sources ? $sources : '[]');
 			try {
 				$this->messageMapper->insert($message);
 			} catch (\OCP\DB\Exception $e) {
--- a/lib/Listener/ChattyLLMTaskListener.php
+++ b/lib/Listener/ChattyLLMTaskListener.php
@@ -61,6 +61,7 @@ class ChattyLLMTaskListener implements IEventListener {
 			$message->setRole('assistant');
 			$message->setContent(trim($task->getOutput()['output'] ?? ''));
 			$message->setTimestamp(time());
+			$message->setSources(json_encode($task->getOutput()['sources'] ?? []));
 			try {
 				$this->messageMapper->insert($message);
 			} catch (\OCP\DB\Exception $e) {
--- a/lib/Service/AssistantService.php
+++ b/lib/Service/AssistantService.php
@@ -74,6 +74,8 @@ class AssistantService {
 		TextToTextTopics::ID => 10,
 	];
 
+	public array $informationSources;
+
 	public function __construct(
 		private ITaskProcessingManager $taskProcessingManager,
 		private TaskNotificationMapper $taskNotificationMapper,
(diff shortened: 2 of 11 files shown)


lib/Service/AssistantService.php
```diff
@@ -86,6 +88,40 @@ public function __construct(
...
+			'list_assignees,' => $this->l10n->t('OpenProject'),
+			'create_work_package' => $this->l10n->t('OpenProject'),
+			'list_talk_conversations' => $this->l10n->t('Nextcloud Talk'),
+			'create_public_conversation' => $this->l10n->t('Nextcloud Talk'),
+			'send_message_to_conversation' => $this->l10n->t('Nextcloud Talk'),
+			'list_messages_in_conversation' => $this->l10n->t('Nextcloud Talk'),
+			'duckduckgo_results_json' => $this->l10n->t('DuckDuckGo Web Search'),
+			'' => $this->l10n->t(''),
+			'' => $this->l10n->t(''),
+			'' => $this->l10n->t(''),
```
julien-nc:
```suggestion
```

src/components/ChattyLLM/Message.vue
```diff
@@ -108,11 +141,13 @@ export default {
 			// with [] we inhibit the extract+resolve mechanism on NcRichText
 			// TODO This can be removed (and all the custom extract+resolve logic) when fixed in NcRichText
 			references: [],
+			sources: [],
 		}
 	},
 
 	mounted() {
 		this.fetch()
+		this.sources = this.parse_sources()
```
julien-nc, line 150: This can be done with a `parsedSources` computed prop (same code as your `parse_sources` method). It's cleaner since the computed will get updated if needed (not needed for now but who knows).

julien-nc (changes requested):

src/components/ChattyLLM/Message.vue
janepie: Thanks, I think I got what you meant

julien-nc (dismissed):

src/components/ChattyLLM/Message.vue
```diff
@@ -111,6 +144,14 @@ export default {
 		}
 	},
 
+	computed: {
+		sources() {
```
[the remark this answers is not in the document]
janepie: Sure, done!

lib/Service/AssistantService.php
```diff
@@ -86,6 +88,36 @@ public function __construct(
...
+			'find_free_time_slot_in_calendar' => $this->l10n->t('Nextcloud Calendar'),
+			'add_task' => $this->l10n->t('Nextcloud Tasks'),
+			'find_details_of_current_user' => $this->l10n->t('Nextcloud User Profile'),
+			'list_decks' => $this->l10n->t('Nextcloud Deck'),
+			'add_card' => $this->l10n->t('Nextcloud Deck'),
+			'get_coordinates_for_address' => $this->l10n->t('Openstreetmap'),
+			'get_current_weather_for_coordinates' => $this->l10n->t('Norwegian Meteorological Institute Weather Forecast'),
+			'get_public_transport_route_for_coordinates,' => $this->l10n->t('HERE Public Transport API'),
+			'get_osm_route,' => $this->l10n->t('Openstreetmap'),
+			'get_osm_link,' => $this->l10n->t('Openstreetmap'),
```
marcelklehr, line 106: Nitpick: It's spelled [OpenStreetMap](https://en.wikipedia.org/wiki/OpenStreetMap) :)

janepie:
> Very nice! Only tiny feedback: In other places where we use the info icon (like in admin settings) it is usually the circular one rather than the square one.

I chose this to make it a bit more eye-catching so people actually click it, but would also be fine with a circular one. Would you prefer that?

lib/Listener/ChattyLLMTaskListener.php
```diff
@@ -61,6 +61,7 @@ public function handle(Event $event): void {
 			$message->setRole('assistant');
 			$message->setContent(trim($task->getOutput()['output'] ?? ''));
 			$message->setTimestamp(time());
+			$message->setSources(json_encode($task->getOutput()['sources'] ?? []));
```
marcelklehr, line 64: json_encode returns false on failure, we might want to check for that

marcelklehr (approved):

lib/Listener/ChattyLLMTaskListener.php
marcelklehr:
```suggestion
			$sources = json_encode($task->getOutput()['sources'] ?? []);
			$message->setSources($sources ? $sources : '[]');
```

I think we want this

PR #232, opened 2025-05-02
feat: update translate task form layout

### LINKED ISSUE #216: Improve the translate UI
For now the translate input form is generated automatically.
It would be nice to have a left-right layout that looks like the custom one implemented in Text.

edward-ly:
Resolves #216. Parts of the new custom translate form component are adapted from the [translate modal](https://github.com/nextcloud/text/blob/main/src/components/Modal/Translate.vue) in the Text app.

Screenshot 2025-05-01 at 23-56-42 Dashboard - Nextcloud


DaphneMuller:
I'm wondering if we want the user to select the model. Didn't we say in the past we let model selection be done by the admin? @julien-nc

janepie:
> I'm wondering if we want the user to select the model. Didn't we say in the past we let model selection be done by the admin? @julien-nc

This is hidden under "show advanced options" right now as we have it in every other task type

janepie:
You seem to not use the input shape defaults right now I think, it would be good to have them used as pre-set languages @edward-ly

edward-ly:
> You seem to not use the input shape defaults right now I think, it would be good to have them used as pre-set languages @edward-ly They should be available as `translateTaskType.inputShapeDefaults`

Thanks for the catch! That should be fixed now.

julien-nc:
Nice! I hope you don't mind: I added a fix to make the output text field show and hide the "copy" button when the value changes. I realized it was broken in the generic form as well because the wrong event was emitted in AssistantFormOutputs so i fixed this one too.

src/components/Translate/TranslateForm.vue
```diff
@@ -0,0 +1,240 @@
...
+
+	.wrapper {
+		display: grid;
+		grid-template-columns: repeat(2, minmax(300px, 1fr));
+		grid-column-gap: 16px;
+
+		.col {
+			grid-row: 1/2;
+		}
+	}
```
julien-nc, line 214:
```suggestion
	.wrapper {
		display: flex;
		gap: 16px;

		.col {
			width: 100%;
		}
	}
```
Flex boxes are simpler to handle IMO.

src/components/Translate/TranslateForm.vue
```diff
@@ -0,0 +1,240 @@
...
+		// don't set the default values if there is a loaded task (initial or from history)
+		if (this.translateTaskId === null) {
+			this.setDefaultValues(false)
+		}
+
+		this.autosize()
+	},
+
+	methods: {
+		autosize() {
```
julien-nc, line 143: Not sure this is necessary. If we never call autosize, both text fields behave like in the generic form, they grow vertically (and independently) until they reach their max height. I think I prefer this. Wdyt?

julien-nc (commented):
2 potential small changes.
Otherwise, good to go :+1:

src/components/Translate/TranslateForm.vue
edward-ly: Good idea, maybe the translate modal in the Text app could use the same boxes too?

src/components/Translate/TranslateForm.vue
julien-nc: My main issue with the current state is that the output field does not grow if the input one contains few lines. But yeah it's kinda fine like it is.

julien-nc (approved):

PR #234, opened 2025-05-06
fix(ChattyLLM): Fix disappearing messages

marcelklehr:


lib/Controller/ChattyLLMController.php
```diff
@@ -283,6 +283,7 @@ public function newMessage(int $sessionId, string $role, string $content, int $t
 			$message->setRole($role);
 			$message->setContent($content);
 			$message->setTimestamp($timestamp);
+			$message->setSources('[]');
```
kyteinsky, line 286: it should be automatically be populated by the default value, no?
https://github.com/nextcloud/assistant/blob/e1ab3907439b8158d64218b22edf4dacee961b43/lib/Migration/Version020500Date20250425125359.php#L35

kyteinsky (approved):

lib/Controller/ChattyLLMController.php
marcelklehr: Feel free to test. I'm not sure why that doesn't work.

marcelklehr: the message gets returned without refetching it from the db. In the db the default is applied, but not in the message object

kyteinsky: ah alright that makes sense. Nice one.

julien-nc (approved):

PR #236, opened 2025-05-07
Open output media list items in the viewer

julien-nc:
It was only done for simple/single output media items. This is now the case for output media list (list of images and files).
The only task type with an output media list is "image generation" for now.

[Kooha-2025-05-07-11-31-48.webm](https://github.com/user-attachments/assets/e22ee3df-302d-495e-86d4-2b0a9e753a4d)

### DIFF
--- a/src/components/fields/ImageDisplay.vue
+++ b/src/components/fields/ImageDisplay.vue
@@ -4,7 +4,9 @@
 -->
 <template>
 	<div class="image-display">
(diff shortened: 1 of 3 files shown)


janepie (approved):
Cool! This translates to the reference widget, right? I should add image generation to context agent I guess

edward-ly (approved):

PR #237, opened 2025-05-07
Change how small number inputs are rendered

julien-nc:
If the field's default value is under 10, we render it like that:

image

refs #215

### DIFF
--- a/src/components/AssistantFormInputs.vue
+++ b/src/components/AssistantFormInputs.vue
@@ -16,6 +16,8 @@
 				:optional-shape="selectedTaskType.optionalInputShape ?? null"
 				:shape-options="selectedTaskType.inputShapeEnumValues ?? null"
 				:optional-shape-options="selectedTaskType.optionalInputShapeEnumValues ?? null"
+				:defaults="selectedTaskType.inputShapeDefaults"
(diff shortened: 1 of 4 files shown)


edward-ly (commented):
Some possible design improvements:
- Add an "active" attribute highlighting the button that was last clicked (or the default value when the form is opened).
- Add a "Custom" button which toggles the visibility of the input field (which would be hidden by default to avoid mixing input types). Clicking on one of the 1-5 buttons would also hide the input field.

julien-nc:
@edward-ly Thanks for the feedback and suggestions. Both are implemented.

It was a bit tricky to make the state reactive to outside changes, especially when typing into the "custom" input field. If you wanna type "11", you start by typing "1" and the value changes immediately and 1 is considered as not custom so the 1 button is highlighted and the input field disappears. Simple solution is to debounce the number input update event.

[Kooha-2025-05-15-11-42-04.webm](https://github.com/user-attachments/assets/c35d777b-cc83-4836-86fe-e2b249b61654)

janepie (approved):
Looks good! Cannot test with >4 pictures right now, because OpenAI has a max of 4 I think. Maybe something we could chunk in the future?

marcelklehr:
Looks good!

PR #246, opened 2025-05-27
Display the assitant modal on top of the smart picker

julien-nc:
We can't control where the smart picker modal is gonna be mounted but we know it is already mounted when displaying the assistant modal. So we can try to move it at a convenient location in the DOM.

Also, this PR sets the container prop of the assistant modal to null to make sure it is mounted where we want an not at the end of `<body>`.

### DIFF
--- a/src/assistant.js
+++ b/src/assistant.js
@@ -44,12 +44,13 @@ window.assistantPollTimerId = null
  * @param {boolean} params.isInsideViewer Should be true if this function is called while the Viewer is displayed
  * @param {boolean} params.closeOnResult If true, the modal will be closed when getting a sync result
  * @param {Array} params.actionButtons List of extra buttons to show in the assistant result form (only if closeOnResult is false)
+ * @param {HTMLElement} params.mountPoint The DOM element in which the assistant modal will be mounted
  * @return {Promise<unknown>}
  */
 export async function openAssistantForm({
 	appId, taskType = null, input = '', inputs = {},
 	isInsideViewer = undefined, closeOnResult = false, actionButtons = undefined,
-	customId = '', identifier = '',
+	customId = '', identifier = '', mountPoint = null,
 }) {
 	const { default: Vue } = await import('vue')
 	const { default: AssistantTextProcessingModal } = await import('./components/AssistantTextProcessingModal.vue')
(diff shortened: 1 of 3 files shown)


kyteinsky:
A very dirty way could be to detect if the modal is opened from a smart picker and instead of opening the viewer, open the file in a new tab.

julien-nc:
@juliusknorr Thanks for reporting this.

We now insert the assistant modal mount point right after `#content` (or `#content-vue`) by default so we can open the viewer on top of it to display output files.

I think the solution in Text is to change the place where the assistant modal is mounted when it's opened from the Text "right" button. There's a new `mountPoint` param accepted by `window.OCA.Assistant.openAssistantForm`. I tried 

``` javascript
const assistantModalMountPoint = document.createElement('div')
assistantModalMountPoint.id = 'assistantModal'
document.querySelector('body').append(assistantModalMountPoint)
```
and pass `mountPoint: assistantModalMountPoint` to `window.OCA.Assistant.openAssistantForm` in `text/src/components/Assistant.vue` and it works fine.

---

But there is a kind of deadlock here.
* If we want to be able to open the viewer from the assistant (to view result files), we need the assistant to be behind the viewer
* If we want to be able to open the assistant while having Text open, we need the assistant to be on top of the (Text) viewer (that's solved by changing where it's mounted), but then we can't open the viewer again to display output files, it will replace the Text viewer.

This is a tough one to tackle. As far as I understand, it's not possible to open 2 viewers at the same time.

juliusknorr:
> This is a tough one to tackle. As far as I understand, it's not possible to open 2 viewers at the same time.

Yes, when you call viewer again while it is open the existing viewer instance will just load the other file. 

Maybe I miss some context here, but what example cases are there where assistant would open a viewer itself? Is that something that is actually useful when you launched assistant from a text document to generate or interact with some selected text?

julien-nc:
For Text, probably not (yet). The only output we're interested in is text.
But if the user plays around in the assistant and tries to "view" an output file in the assistant, the viewer opens and replaces Text.

Are you fine with my suggested solution to mount the assistant modal somewhere else?

juliusknorr:
Yes, that seems reasonable, then we may need the same in richdocuments.

Besides that maybe we should discuss at some point also if we want the full assistant within those editing tools or maybe some limited feature set only allowing to interact within the current context.

julien-nc:
Indeed, we mentioned that a long time ago and nothing happened.

PR #253, opened 2025-06-05
Make sure the assistant is mounted after the existing viewer

julien-nc:
We can already know if the assistant should be displayed on top of an existing viewer. The `isInsideViewer` of `openAssistantTask()` and `openAssistantForm()` tells us that.

So instead of asking integrators to specify a mount point for that case, we can automatically mount the assistant at the end of the `<body>`.

The `openAssistantTask()` function has been adjusted to be consistent with `openAssistantTask()`.


juliusknorr:
CI run to see if the failing text test passes again: https://github.com/nextcloud/text/pull/7286

PR #255, opened 2025-06-10
fix: assistant fails to load when text2text disabled but other task exists

lukasdotcom:
Loading spinner will spin forever


julien-nc (approved):
Looks good :+1:

PR #256, opened 2025-06-11
feat: show loading screen in task area only

### LINKED ISSUE #245: Don't block the screen with the loading view
Right now the loading view (EmptyContent) is replacing the entire content of the assistant.
It could just replace the input/output form (the "right side"/"main content" of the sidebar layout).

* The "back to the assistant" button is not needed anymore
* The "Run task in the background and get notified" can stay and be renamed to "Get notified when the task finishes". The icon can be changed to the notification one (the bell outline).
* The "Cancel task" button can stay

The UI flow would be:

* When the selected task's state is scheduled/running, we display the EmptyContent instead of the input/output form
* When the EmptyContent is visible, we poll the task's state
* When switching to another task (click on a task in the left sidebar)
    * we stop polling the task
    * we get the state of the new selected task
    * if the state is scheduled/running, we show the loading view and start polling this task
* When clicking on the "Cancel task" button, stop polling, send the cancel request, load the task state again
* When clicking on the "get notified" button, the icon could change to a full/solid bell

When showing the loading view, there could be one request to the assistant backend to know if a notification has been requested and choose the outline or plain bell icon accordingly.
GET `/api/{apiVersion}/task/{ocpTaskId}/notify`

edward-ly:
Fixes #245. In addition, the task list is automatically updated when a new task is submitted or a running task is canceled, with the submitted task also being automatically selected from the list.


julien-nc (changes requested):
This looks great! A few things to adjust and we're good to go.

The task list item is not updated when the task finishes. Maybe we should fire an event in the event bus after pollTask so the TaskList component can update its internal task object status.
`emit('task-updated', task)`
which would be caught in TaskList and call a method like:
``` js
updateTask(task) {
    const internalTask = this.tasks.find(t => t.id === task.id)
    if (internalTask) {
        internalTask.status = task.status
    }
}
```

edward-ly:
> The task list item is not updated when the task finishes. Maybe we should fire an event in the event bus after pollTask so the TaskList component can update its internal task object status. `emit('task-updated', task)` which would be caught in TaskList and call a method like:
> 
> ```js
> updateTask(task) {
>     const internalTask = this.tasks.find(t => t.id === task.id)
>     if (internalTask) {
>         internalTask.status = task.status
>     }
> }
> ```

Is `emit('task-updated', task)` being called in `assistant.js` when the task finishes or somewhere else? How can the TaskList component listen to an emit from a grandparent?

lib/Service/AssistantService.php
```diff
@@ -153,6 +182,33 @@ public function notifyWhenReady(int $taskId, string $userId): void {
...
+	 * @param int $taskId
+	 * @param string $userId
+	 * @throws Exception
+	 * @throws MultipleObjectsReturnedException
+	 */
+	public function cancelNotifyWhenReady(int $taskId, string $userId): void {
+		try {
+			$task = $this->taskProcessingManager->getTask($taskId);
+		} catch (NotFoundException $e) {
+			// task may already be deleted, nothing to do
```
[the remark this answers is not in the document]
edward-ly: done in https://github.com/nextcloud/assistant/pull/256/commits/c96d434a904ddfa7b84b81394891a84c18cc971f

src/assistant.js
```diff
@@ -154,57 +154,56 @@ export async function openAssistantForm({
...
+						view.selectedTaskId = updatedTask.id
+						lastTask = updatedTask
+						return
+					}
 
-						view.loading = true
-						view.showSyncTaskRunning = true
-						view.progress = null
-						view.expectedRuntime = (updatedTask?.completionExpectedAt - updatedTask?.scheduledAt) || null
+					view.loading = true
```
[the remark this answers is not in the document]
edward-ly: done in https://github.com/nextcloud/assistant/pull/256/commits/e5a341999e75aa34399ea95ee7600a7a19c9f556, thanks for the reminder

lib/Service/AssistantService.php
```diff
@@ -153,6 +182,32 @@ public function notifyWhenReady(int $taskId, string $userId): void {
...
+		try {
+			$task = $this->taskProcessingManager->getTask($taskId);
+		} catch (NotFoundException $e) {
+			// ignore, task may be already deleted, so delete any dangling notifications
+		} catch (TaskProcessingException $e) {
+			$this->logger->debug('Task request error : ' . $e->getMessage());
+			throw new Exception('Internal server error.', Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+
+		if ($task->getUserId() !== $userId) {
```
julien-nc, line 203: `$task` might notbe defined here.

lib/Service/AssistantService.php
```diff
@@ -153,6 +182,32 @@ public function notifyWhenReady(int $taskId, string $userId): void {
...
+	 * @param int $taskId
+	 * @param string $userId
+	 * @throws Exception
+	 * @throws MultipleObjectsReturnedException
+	 */
+	public function cancelNotifyWhenReady(int $taskId, string $userId): void {
+		try {
+			$task = $this->taskProcessingManager->getTask($taskId);
+		} catch (NotFoundException $e) {
+			// ignore, task may be already deleted, so delete any dangling notifications
```
julien-nc, line 197:
```suggestion
			// ignore, task may be already deleted, so delete any dangling notifications
			$this->taskNotificationMapper->deleteByTaskId($taskId);
			return;
```

src/assistant.js
```diff
@@ -213,19 +219,22 @@ export async function openAssistantForm({
...
+			setNotifyReady(lastTask.id, false).finally(res => {
+				view.isNotifyEnabled = false
+				cancelTask(lastTask.id).then(res => {
+					view.showSyncTaskRunning = false
+					view.loading = false
+					view.selectedTaskId = null
+					lastTask = null
+				})
+			})
 		})
```
julien-nc, line 238:
```suggestion
			setNotifyReady(lastTask.id, false)
			cancelTask(lastTask.id).then(res => {
				view.showSyncTaskRunning = false
				view.loading = false
				view.selectedTaskId = null
				lastTask = null
			})
		})
```

We don't need to wait for setNotifyReady to be done and then cancel the task. We can do the requests concurrently.

julien-nc:
> Another alternative could be instead of declaring the tasks list in the TaskList component data, declare it in assistant.js instead and pass it down to TaskList as a component property.

It's a lot of chaining props to carry the value from assistant.js to TaskList. I would be in favor of using the event bus.

When using this function: `import { emit } from '@nextcloud/event-bus'` you can emit an event+data in the event bus.
You can subscribe to this event anywhere, this is independent from Vue.
In TaskList you can use `import { subscribe } from '@nextcloud/event-bus'` to subscribe to the event. You can do so in the `mounted` function:
`subscribe('task-updated', this.updateTask)`
and you need to unsubscribe when the TaskList component is destroyed, in the `beforeDestroy` block:
`unsubscribe('task-updated', this.updateTask)`

lib/Service/AssistantService.php
edward-ly: In which case, delete the task notification anyway? Adding `$task !== null` might be enough in that case.

edward-ly:
> When using this function: `import { emit } from '@nextcloud/event-bus'` you can emit an event+data in the event bus. You can subscribe to this event anywhere, this is independent from Vue. In TaskList you can use `import { subscribe } from '@nextcloud/event-bus'` to subscribe to the event. You can do so in the `mounted` function: `subscribe('task-updated', this.updateTask)` and you need to unsubscribe when the TaskList component is destroyed, in the `beforeDestroy` block: `unsubscribe('task-updated', this.updateTask)`

done in https://github.com/nextcloud/assistant/pull/256/commits/08c1af13c775787965e0c8334b6826c6adab5767

janepie:
Nice one!
One thing: when I submit a task and then switch tasks, I stay in the loading screen and cannot leave it in any way, even the buttons do not work anymore

edward-ly:
> Nice one! One thing: when I submit a task and then switch task type, I stay in the loading screen and cannot leave it in any way, even the buttons do not work anymore

Good catch, should be fixes in 12ecdad now, mind testing one last time? @janepie @julien-nc

src/assistant.js
```diff
@@ -154,78 +156,89 @@ export async function openAssistantForm({
...
+				}).catch(error => {
+					console.error(error)
+				})
 			}
 		})
 		view.$on('new-task', () => {
 			console.debug('[assistant] new task')
+			cancelTaskPolling()
+			view.showSyncTaskRunning = false
+			view.isNotifyEnabled = false
```
julien-nc, line 223: This is needed in `openAssistantTask()` as well.

julien-nc (approved):
The issue with switching task types while a task is "loading" is fixed when opening the assistant with the menu entry but not when opening it from a notification. See inline comment.

src/assistant.js
edward-ly: Is it possible to create a notification for a task that is still running/scheduled? I thought that shouldn't be the case.

edward-ly:
> I still have the loading spinner, I just made a recording

Ah I see, I fixed the case for when a task from the task list is loaded, but not when switching task types or creating a new task, should be fixed now.

janepie (approved):
Looks good, really nice work!

PR #257, opened 2025-06-11
Remove dialog header and duplicate title

marcoambrosini:
If we have enough similar instances we should probably upstream these changes to the NcModal component, and provide a prop to hide the header there.

| Before | After |
|--------|-------|
| Screenshot 2025-06-11 at 11 50 43 | Screenshot 2025-06-11 at 11 52 11 |

### DIFF
--- a/src/components/AssistantTextProcessingModal.vue
+++ b/src/components/AssistantTextProcessingModal.vue
@@ -228,4 +228,13 @@ export default {
 		height: 100%;
 	}
 }
+
+// TODO: upstream the following by adding a `HideHeader` prop to the NcModal component
+:deep(.modal-header) {
+	display: none !important;
+}
+:deep(.modal-container) {
+	top: 0 !important;
+	height: 100% !important;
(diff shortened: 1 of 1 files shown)


julien-nc:
It looks nicer indeed.
I didn't check why yet but: The close button cannot be clicked when the assistant is opened with the smart picker.

julien-nc:
You can enable the "testing" app which contains some providers.

Yes it is a regression. The close button can be clicked when building the main branch.

julien-nc:
Yes it's mounted inside the smart picker modal (which contains nothing, it's a trick) to make sure it is displayed on top of it.

marcoambrosini:
Oh, so that's the header of the smart picker modal :)

janepie:
@marcoambrosini the linked PR was closed, I saw because it was redundant? Does this one still work without it?

marcoambrosini:
This issue has been addressed so we can close this!

PR #258, opened 2025-06-12
Migrate to Vue 3 and nextcloud/vue 9

julien-nc:
It was much longer than expected. There's a lot of things to adjust.

* `prop.sync` -> `v-model:prop`
* many components now take their value as `modelValue`
* fixed an issue when creating a new message in the chattyUI, it crashed because the sources attribute was not defined in the new message
* fixed bug when trying to preserve the "input" field when its type is not "text"
* fixed admin settings initial state: provide the default system prompts if they are empty
* I couldn't find a way to change prop values from the outside (in assistant.js) so we now use data props. We just set the initial values with props when creating the vue app. We can then change data prop values with the root component instance :weary: 
* `view.$destroy()` -> `app.unmount()`
* We can't catch events from outside Vue anymore. So we fire classic events from the modal element and catch them with `elem.addEventListener()`
* NcButton `type` prop is now `variant`
* syntax rules for `:deep` have changed
* replace vue-clipboard2 by `navigator.clipboard.writeText()`

I checked:
* submitting
* polling
* loading a task from the history
* opening a task's results from a notification
* smart picker
* text integration
* opening an output file with the viewer

It would be nice to play around and try everything. For example the agency feature.

### DIFF
--- a/src/components/AssistantTextProcessingForm.vue
+++ b/src/components/AssistantTextProcessingForm.vue
@@ -34,7 +34,8 @@
 							:selected-task-id="selectedTaskId"
 							class="history--list"
 							@try-again="onHistoryTryAgain"
-							@load-task="onHistoryLoadTask" />
+							@load-task="onHistoryLoadTask"
+							@task-deleted="onHistoryTaskDeleted" />
 					</NcAppNavigationList>
 				</NcAppNavigation>
 				<RunningEmptyContent
@@ -504,6 +505,13 @@ export default {
 
 			this.$refs?.translateForm?.setDefaultValues(true)
 			this.$refs?.assistantFormInputs?.setDefaultValues(true)
+		},
+		onHistoryTaskDeleted(task) {
+			// if the currently selected task is deleted, simulate a click on "new task"
+			// this will stop the task status polling and display the form again
+			if (task.id === this.selectedTaskId) {
+				this.onHistoryNewTask()
+			}
 		},
 	},
 }
--- a/src/components/TaskList.vue
+++ b/src/components/TaskList.vue
@@ -69,6 +69,7 @@ export default {
 		'load-task',
 		'try-again',
 		'update:loading',
+		'task-deleted',
 	],
 
 	data() {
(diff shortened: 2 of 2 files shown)


julien-nc:
> The action confirmation thingie doesn't respect chosen colors completely, which looks a bit weird

Hmmm try with the main branch. It looks exactly the same for me in `main` and `enh/noid/vue3`:

image

The note card is blue, that does not change with the theming options. The buttons follow theming and Agency actions has `background-color: var(--color-primary-element-light-hover);`. It's all intended. Do you have suggestions to improve this? Let's do it in another PR though.

> Dropdown menus don't work properly, I cannot choose anything

Done, thanks. Some NcCheckbox were not working as well. All fixed.

> Run task in the background and get notified' button doesn't work

And "Cancel" didn't work either. Bother are fixed.

julien-nc:
Thanks for the review :pray:

janepie:
Will open a new issue on the note card color :)

Two more things: when choosing a file for Transcribe audio

- [ ] from my device: when I press Cancel, the Assistant modal closes
- [ ] from Nextcloud: the modal behaves weird and has a weird sizing

janepie:
Okay, just checked, that part is also happening on main, so we can ignore it for now (and I guess it's not on our side). But the Choose button lost its color

julien-nc:
Hey, I rebased on main and solved a lot of conflicts. Could you check again that everything is working?

julien-nc:
One more fix: the `can-close` prop of `NcModal` became `no-close`.

julien-nc:
I spotted an extra issue (not related with Vue 3):
When the currently selected task is deleted (via the history entry context menu), the form is not updated so:
* If the task is scheduled/running: The loading screen stays and we keep polling the task state
* If the task is done/canceled/anything else: The form keeps showing the task result

Solution: When the task is actually deleted (in `<TaskList>`), let the upper context know (emit event caught by `<AssistantTextProcessingForm>`) and create a new task (simulate a click on "new task") so the form is cleared and we stop polling.

I'll make more checks and merge that soon.

PR #262, opened 2025-06-17
Fix input preservation

julien-nc:
Preserve inputs when switching task type only if it was text and it goes in a text field.


janepie:
like a charm ❤️

janepie (approved):

PR #288, opened 2025-06-27
Use outline icons where it makes sense

julien-nc:
And replace custom icons by material ones.


edward-ly:
also ShareVariantIcon to ShareVariantOutlineIcon, maybe

kyteinsky (commented):
Not really related to vue 3 migration but it could be a good place to fix this. It happens when the text contains large texts that cannot be wrapped on spaces like long URLs. The browser window has been shrunk horizontally too.

image

These seem like vue 3 regression/prop change.
1. User status is disabled so should not be displayed for "admin". The new prop is `hideStatus` .
2. Assistant icon is not centered

image

julien-nc:
@kyteinsky Thanks. I think both those issues are now fixed (with the last 2 commits).

PR #295, opened 2025-07-07
fix: explicitly set default sources when creating new session

hamza221:
Extracts from nextcloud.log
```
 "Exception": "OC\\DB\\Exceptions\\DbalException",
    "Message": "An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1364 Field 'sources' doesn't have a default value",

 "message": "Failed to create a chat session",
    "exception": {
      "query": "INSERT INTO `*PREFIX*assistant_chat_msgs` (`session_id`, `role`, `content`, `timestamp`) VALUES(:dcValue1, :dcValue2, :dcValue3, :dcValue4)"
    },
    "CustomMessage": "Failed to create a chat session"
```

### DIFF
--- a/lib/Controller/ChattyLLMController.php
+++ b/lib/Controller/ChattyLLMController.php
@@ -147,6 +147,7 @@ class ChattyLLMController extends OCSController {
 			$systemMsg->setRole('system');
 			$systemMsg->setContent($userInstructions);
 			$systemMsg->setTimestamp($session->getTimestamp());
+			$systemMsg->setSources('[]');
 			$this->messageMapper->insert($systemMsg);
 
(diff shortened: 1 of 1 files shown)


kyteinsky (approved):
thanks Hamza!

julien-nc (approved):
Thank you

julien-nc:
I can't reproduce that, the value value is effective in my dev setups. So it seems the default value is not set in the context where it fails. I would be curious to know if it's an isolated issue or a systematic one with some conditions.

janepie (approved):
Thanks for the fix! 
I'd also still be interested why this is happening in the first place

Issue #301, opened 2025-07-09
Reduce total amount of width of the assistant


marcoambrosini:
## Problem

The assistant is currently full width, which on big screens affects readability and usability

## Agreed solution
Use the "full" modal but clamp the width of the content at the max-width of 1200px. The textareas in the translate tab should be higher by default to make it looks less strange. Instead of the current ~3 lines, just make it 10 lines or so.

DaphneMuller:
What is the reason for this? In the previous design the modal was smaller and then we got the request from the design team to make it fullscreen. So this feedback is not aligned with previous feedback

julien-nc:
Here is how it looks with a "large" modal: 

<img width="1915" height="921" alt="Image" src="https://github.com/user-attachments/assets/65b07489-0eee-495c-965a-f300dc8be417" />

Daphne and me think it's not so nice and the full width is still better.

For comparison, here is the current modal size: 

<img width="1915" height="921" alt="Image" src="https://github.com/user-attachments/assets/ed326d6c-05ce-4c4d-b3c0-dec226da1da7" />

julien-nc:
The NcModal component has no intermediate width between "large" and "full".

DaphneMuller:
@marcoambrosini can you provide a decision if you would really like to move forward with the adjustment

marcoambrosini:
I think 1200 would be better, especially considering the translate view which has two columns.

marcoambrosini:
<img width="2239" height="1221" alt="Image" src="https://github.com/user-attachments/assets/cd904a75-1269-451b-9f5a-5185c72758b7" />

This is how the translate tab looks clamped at 1200

Issue #300, opened 2025-07-09
Add descriptive icons to the assistant tabs


marcoambrosini:

DaphneMuller:
We will try to assign someone to this as soon as possible

DaphneMuller:
@marcoambrosini could you provide which icons we should use for which feature?

marcoambrosini:
Here it goes:

* Chat with AI - https://fonts.google.com/icons?selected=Material+Symbols+Outlined:chat_bubble:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=chat&icon.size=24&icon.color=%235f6368
* Work with text - https://fonts.google.com/icons?selected=Material+Symbols+Outlined:subject:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=subject&icon.size=24&icon.color=%235f6368
* Work with images - https://fonts.google.com/icons?selected=Material+Symbols+Outlined:image:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=image&icon.size=24&icon.color=%235f6368
* Translate - https://fonts.google.com/icons?selected=Material+Symbols+Outlined:language:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=translate&icon.size=24&icon.color=%235f6368
* Work with audio - https://fonts.google.com/icons?selected=Material+Symbols+Outlined:graphic_eq:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=waveform&icon.size=24&icon.color=%235f6368
* Generate file - https://fonts.google.com/icons?selected=Material+Symbols+Outlined:draft:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=file&icon.size=24&icon.color=%235f6368
* Other - https://fonts.google.com/icons?selected=Material+Symbols+Outlined:more_horiz:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=more&icon.size=24&icon.color=%235f6368

Issue #299, opened 2025-07-09
Simplify and group Assistant tabs


marcoambrosini:
Possible configuration that would bring this down to 6 tabs

## * Work with text: 
  * Generate text
  * Context write
  * Generate a headline
  * Summarize
  * Extract topics
  * Proofread
  * Reformulate text
  * Change tone

## * Work with images
  * Generate images

## * Work with audio
  * Transcribe audio
  * Text to speech

## * Generate documents
  * Generate office document
  * Generate spreadsheet
  * Generate presentation

## * Other
  * For unsorted apps

<img width="1896" height="4451" alt="Image" src="https://github.com/user-attachments/assets/2266239a-c665-4d1f-874e-ef6f6dbd2af9" />

There should be icons on the tabs: https://github.com/nextcloud/assistant/issues/300

the main tabs would unfold when clicked to show all the task types of the category.

jancborchardt:
Some feedback:
- This is a very technical way of sorting things, thinking about "input" and "output". Try and think about it more of how it would be grouped from a user perspective.
- "Text + file input" is much too large of a group
- "Generate image" has to be a distinct entry. It’s a common case that from a user perspective has little to do with "file/text". It could also be combined into an "Images" group with emoji generation and analyzing images.
- There can be a group "Generate document" with: Generate office document, Generate spreadsheet, Generate presentation
- Another possible group is "Modify text" (wording can be adjusted). This is basically what would be shown in the menu when you select text in Nextcloud Text. With following entries:
   - Summarize
   - Generate a headline
   - Extract topics
   - Reformulate text
   - Change tone
   - Proofread
- It’s good to keep translation separate, not only because of the layout difference, but because people don’t think of it as "text input" or "modify text".

julien-nc:
After discussion with @marcoambrosini we decided that the main 6 tabs would unfold when clicked to show all the task types of the category. Are we on the same page?

janepie:
Sure, just wanted to align if we want to have an "Other" or an extra tab for every unsorted one. To be able to sort them if they fit, we'd need some way to classify the types on the TaskType side, not hardcoded in Assistant

DaphneMuller:
Just for context, the "other" one will be mostly used for apps developed by community, as we have control over the apps we develop

marcelklehr:
Something to think about is that with agency being able to run most of the tasks directly inside the chat, we might be able to get rid of some tabs. What do you think?

marcoambrosini:
[...]

PR #311, opened 2025-07-14
feat: Add categories with icons

lukasdotcom:
Resolves: #299
Resolves: #300
Resolves: #301
image

### Some design questions.

1. For some reason the waveform icon is different in the vue material icons than on the material icons (or I couldn't find it🤷) so had to upload the icon from google fonts. The one that vue material icons comes with is: image

2. I'm not sure if context chat should have its own category, but left it in text (can be moved to other). EDIT: Added category for context chat with icon that was discussed with @marcoambrosini 

3. I changed the maximum width to 1220px instead of 1200px to ensure that all categories can fit on one line in english.

### Other notes.
1. The rules to assign categories are built to be flexible and guess categories for new tasks as long as they follow some implied rules. 
2. I did remove the inline menu, but theoretically could add the old one back just doesn't work well with categories.
3. Also added convenience feature that if only one task exists in category clicking on the category will bring you to that task.

### DIFF
--- a/src/components/TaskTypeSelect.vue
+++ b/src/components/TaskTypeSelect.vue
@@ -5,28 +5,40 @@
 <template>
 	<div ref="taskTypeSelect"
 		class="task-type-select">
-		<NcActions v-for="variants in buttonTypesByInlineStatus.inline"
-			:key="variants.id"
-			:force-menu="true"
-			:menu-name="variants.text"
-			:container="$refs.taskTypeSelect"
-			:primary="selectedCategory(variants)"
-			@click="onMenuCategorySelected(variants.id, variants.tasks)">
-			<NcActionButton v-for="t in variants.tasks"
-				:key="t.id"
-				:disabled="selectedTask(t)"
-				:title="t.description"
-				:close-after-click="true"
-				@click="onTaskSelected(t)">
+		<template v-for="variants in buttonTypesByInlineStatus.inline">
+			<NcActions v-if="hasSubMenu(variants)"
+				:key="variants.id"
+				:force-menu="true"
+				:menu-name="variants.text"
+				:container="$refs.taskTypeSelect"
+				:primary="isCategorySelected(variants)"
+				@click="onMenuCategorySelected(variants)">
+				<NcActionButton v-for="t in variants.tasks"
+					:key="t.id"
+					:disabled="selectedTask(t)"
+					:title="t.description"
+					:close-after-click="true"
+					@click="onTaskSelected(t)">
+					<template #icon>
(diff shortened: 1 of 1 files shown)


marcelklehr:
I'd be in favor of moving context chat out of the Working with text category

lukasdotcom:
The last commit added an overflow menu that allows you to select categories and tasks. Note that the amount of inlined tasks needs to be specified as a prop currently by the parent.
image

src/components/TaskTypeSelect.vue
```diff
@@ -78,66 +115,137 @@ export default {
...
-				this.extraButtonType = selectedAction
-			}
-		},
-		getButtonType(taskType) {
+		selectedTask(taskType) {
 			return taskType.id === this.modelValue
-				? 'primary'
-				: 'secondary'
+		},
+		selectedCategory(category) {
```
julien-nc, line 180:
```suggestion
		isCategorySelected(category) {
```
More explicit IMO

src/components/TaskTypeSelect.vue
```diff
@@ -78,66 +115,137 @@ export default {
...
-				const types = this.options.slice(this.inline).filter(t => t.id !== this.modelValue && t.id !== this.extraButtonType.id)
-				// add the one that was a button and that has been replaced
-				if (this.extraButtonType.id !== this.options[this.inline - 1].id) {
-					types.unshift(this.options[this.inline - 1])
+		buttonTypesByInlineStatus() {
+			if (this.onlyInline) {
+				return { inline: this.buttonTypes, overflow: [] }
+			}
+			const inline = this.buttonTypes.slice(0, this.inline)
+			let overflow = this.buttonTypes.slice(this.inline)
```
julien-nc, line 154:
```suggestion
			const inlineButtonTypes = this.buttonTypes.slice(0, this.inline)
			let overflowButtonTypes = this.buttonTypes.slice(this.inline)
```
Maybe?

julien-nc:
Also, I would recommend to rebase on main from time to time. A small conflict is better than a big conflict :grin:

lukasdotcom:
> A bit of nitpicking 😁 Also, could we prevent opening the actions dropdown when clicking on a category in which there is only one task type? Maybe we should only do that when the category name equals the single task type name. It's the case for translate and chat with AI.

I changed it to create a NcButton when there is no dropdown needed because I couldn't find a way to hide it and implemented the other minor name changes.

jancborchardt (commented):
Looks really good @lukasdotcom! :) It will help people a lot to have a better overview

PR #312, opened 2025-07-15
Add file actions

### LINKED ISSUE #304: Files actions using TaskProcessing

There could be file actions to process files and produce new files. For example:
* Summarize a text file to a new text file
* Transcribe an audio file to a new text file
* Text-to-speech a text file to a new audio file

cc @jancborchardt 

Register file actions in the assistant.

No alternative 😁

julien-nc:
closes #304

3 File actions are registered:
* summarize for text and markdown files
* STT for audio files
* TTS for text and markdown files

A notification is sent after a file action process finishes or fails.

Main changes:
* Restructure the notifier
* New endpoint in the API controller to run the actions
* 2 new listeners for TaskFailed and TaskSuccess to handle tasks related with file actions


julien-nc:
@marcoambrosini @jancborchardt I made a first implementation. Let's adjust it a bit.
Right now, for text files and wav/mp3 files, we have a new file action group:
image
For text files:
image
For audio files:
image

Once it is clicked, there's a top-right toast saying "A file will be created soon".
Then the task is scheduled and when it's done, a new file is created next to the source one with a suffix (_summarized, _tts or _stt).

Questions:
* Are you fine with the action names and icons?
* Should we adjust the toast msg?
* Are you fine with the behaviour or creating a file next to the source one?
* Are you fine with the suffix?
* Should there be a notification when the new file is created?

jancborchardt (commented):
Awesome awesome @julien-nc! :) 


> Questions:
> 
>     * Are you fine with the action names and icons?

The names are good! Best to use the Material Symbols for it:
- Summarize: [Looks good](https://fonts.google.com/icons?selected=Material+Symbols+Outlined:summarize:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=summariz&icon.size=24&icon.color=%23e3e3e3)
- Text to speech: https://fonts.google.com/icons?selected=Material+Symbols+Outlined:speech_to_text:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=transcri&icon.size=24&icon.color=%23e3e3e3
- Transcribe audio: Better icon → https://fonts.google.com/icons?selected=Material+Symbols+Outlined:speech_to_text:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=transcri&icon.size=24&icon.color=%23e3e3e3


>     * Should we adjust the toast msg?

Can you check for the wording @marcoambrosini? It's about the showSuccess and showError texts I assume.


>     * Are you fine with the behaviour or creating a file next to the source one?
>     * Are you fine with the suffix?

Sure, seems great to put it next to it. Just the naming should be adjusted to be human readable:
- " - summarized"
- " - text to speech"
- " - transcribed"


>     * Should there be a notification when the new file is created?

Good question, I guess you should be notified when it's done just like the Talk recording.


And one more thing: What about direct transcription + summarization? Would be a nice feature too and could just chain the 2 functions in the back?

julien-nc:
@jancborchardt Thanks for the feedback.

* Target file names have been adjusted
* There is now a notification when a file action succeeds and fails (because the task failed or because something else failed). The notifications looks like that: 
image
The source file is a rich object (/f/ link), the action button is also a /f/ link to the target file.
* Material symbols are now used (see updated screenshots in previous comment)

> And one more thing: What about direct transcription + summarization? Would be a nice feature too and could just chain the 2 functions in the back?

I don't have time for that these days. We can keep it as a followup task. (Maybe it's worth creating another GH issue for that).

src/files/fileActions.js
```diff
@@ -0,0 +1,164 @@
...
+]
+
+function registerGroupAction() {
+	const groupAction = new FileAction({
+		id: 'assistant-group',
+		displayName: (nodes) => {
+			return t('assistant', 'Assistant')
+		},
+		enabled(nodes, view) {
+			return !actionIgnoreLists.includes(view.id)
```
lukasdotcom, line 26: This should probably be disabled if none of the assistant tasks are available.

src/files/fileActions.js
```diff
@@ -0,0 +1,164 @@
...
+		parent: 'assistant-group',
+		displayName: (nodes) => {
+			return t('approval', 'Text to speech')
+		},
+		enabled(nodes, view) {
+			return !actionIgnoreLists.includes(view.id)
+				&& nodes.length === 1
+				&& !nodes.some(({ permissions }) => (permissions & Permission.READ) === 0)
+				&& nodes.every(({ type }) => type === FileType.File)
+				&& nodes.every(({ mime }) => ['text/plain', 'text/markdown'].includes(mime))
```
lukasdotcom, line 88: Same thing here as for summarize

src/files/fileActions.js
```diff
@@ -0,0 +1,164 @@
...
+		parent: 'assistant-group',
+		displayName: (nodes) => {
+			return t('approval', 'Transcribe audio')
+		},
+		enabled(nodes, view) {
+			return !actionIgnoreLists.includes(view.id)
+				&& nodes.length === 1
+				&& !nodes.some(({ permissions }) => (permissions & Permission.READ) === 0)
+				&& nodes.every(({ type }) => type === FileType.File)
+				&& nodes.every(({ mime }) => ['audio/mpeg', 'audio/wav', 'audio/mp3'].includes(mime))
```
lukasdotcom, line 123: Probably also worth it to allow transcripts to be generated for video files as that should already be supported.

lib/Listener/FileActionTaskSuccessfulListener.php
```diff
@@ -0,0 +1,101 @@
...
+					$extension = $mimes->getExtension($mimeType);
+					if ($extension === 'mpga') {
+						$extension = 'mp3';
+					}
+					$targetFileName = $sourceFile->getName() . ' - text to speech.' . $extension;
+					$targetFile = $sourceFileParent->newFile($targetFileName, $speechFile->fopen('rb'));
+				} else {
+					$textResult = $task->getOutput()['output'];
+					$suffix = $taskTypeId === TextToTextSummary::ID ? 'summarized' : 'transcribed';
+					$targetFileName = $sourceFile->getName() . ' - ' . $suffix . '.txt';
```
lukasdotcom, line 80: Might be worth it to make it a markdown file instead of text as most AI's like to output text in markdown format.

src/files/fileActions.js
```diff
@@ -0,0 +1,164 @@
...
+		parent: 'assistant-group',
+		displayName: (nodes) => {
+			return t('approval', 'Summarize')
+		},
+		enabled(nodes, view) {
+			return !actionIgnoreLists.includes(view.id)
+				&& nodes.length === 1
+				&& !nodes.some(({ permissions }) => (permissions & Permission.READ) === 0)
+				&& nodes.every(({ type }) => type === FileType.File)
+				&& nodes.every(({ mime }) => ['text/plain', 'text/markdown'].includes(mime))
```
lukasdotcom, line 53: Might be worth it to support all files that parseTextFromFile supports instead of just markdown and plain text.
https://github.com/nextcloud/assistant/blob/0051c8e06ea97bba53cd5522bbdf5c91b9751f03/lib/Service/AssistantService.php#L655-L725

lukasdotcom (changes requested):
Looks good. Tested it out and it worked.
Just a nitpick and some things that should be discussed.

src/files/fileActions.js
julien-nc: Same here, instead of toggling the action with its `enabled` attr, we conditionally register it at the end of the file.

src/files/fileActions.js
julien-nc: As far as I know we don't have a video transcription task type.

src/files/fileActions.js
julien-nc: Makes sense!
I created an issue to keep track of that: #321
We can do it later, after having merged this one.

src/files/fileActions.js
```diff
@@ -0,0 +1,164 @@
...
+		order: 0,
+		async exec(node) {
+			const { default: axios } = await import('@nextcloud/axios')
+			const { generateOcsUrl } = await import('@nextcloud/router')
+			const { showError, showSuccess } = await import('@nextcloud/dialogs')
+			const url = generateOcsUrl('/apps/assistant/api/v1/file-action/{fileId}/core:text2text:summary', { fileId: node.fileid })
+			try {
+				const response = await axios.post(url)
+				console.debug('taskId', response.data.ocs.data.taskId)
+				showSuccess(t('assistant', 'The result file will be created soon'))
```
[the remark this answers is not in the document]
julien-nc: Solved by using "assistant" as task's appId.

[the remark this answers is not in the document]
julien-nc: Done for the 3 file actions.

lib/Notification/Notifier.php
```diff
@@ -184,6 +186,85 @@ public function prepare(INotification $notification, string $languageCode): INot
...
+
+				$notification
+					->setParsedSubject($subject)
+					->setRichMessage($message, [
+						'file' => [
+							'type' => 'file',
+							'id' => (string)$params['source_file_id'],
+							'name' => $params['source_file_name'],
+							'path' => $params['source_file_path'],
+							'link' => $this->url->linkToRouteAbsolute('files.viewcontroller.showFile', ['fileid' => $params['source_file_id']]),
```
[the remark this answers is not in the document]
julien-nc: Done

julien-nc:
@kyteinsky Thanks for the review. All done.

PR #323, opened 2025-08-04
feat: add sticker generation smart picker

lukasdotcom:
Not sure if the sticker prompt should be hidden or not. Currently it is visible. An example sticker if the user adds dog:
image

### DIFF
--- a/src/stickerGeneration.js
+++ b/src/stickerGeneration.js
@@ -1,0 +1,31 @@
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { registerCustomPickerElement, NcCustomPickerRenderResult } from '@nextcloud/vue/components/NcRichText'
+
+registerCustomPickerElement('assistant_sticker_generation', async (el, { providerId, accessible }) => {
+	const { createApp } = await import('vue')
+	const { default: ImageResultCustomPickerElement } = await import('./views/ImageResultCustomPickerElement.vue')
+
+	const app = createApp(
+		ImageResultCustomPickerElement,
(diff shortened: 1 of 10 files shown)


src/stickerGeneration.js
```diff
@@ -0,0 +1,31 @@
...
+
+registerCustomPickerElement('assistant_sticker_generation', async (el, { providerId, accessible }) => {
+	const { createApp } = await import('vue')
+	const { default: ImageResultCustomPickerElement } = await import('./views/ImageResultCustomPickerElement.vue')
+
+	const app = createApp(
+		ImageResultCustomPickerElement,
+		{
+			inputs: {
+				input: 'cartoon, neutral background, sticker of ',
```
julien-nc, line 16: This should be a translated string.

julien-nc (changes requested):
Well done! Works like a charm.

src/stickerGeneration.js
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+import { registerCustomPickerElement, NcCustomPickerRenderResult } from '@nextcloud/vue/components/NcRichText'
+
+registerCustomPickerElement('assistant_sticker_generation', async (el, { providerId, accessible }) => {
+	const { createApp } = await import('vue')
+	const { default: ImageResultCustomPickerElement } = await import('./views/ImageResultCustomPickerElement.vue')
+	console.debug(t('assistant', 'cartoon, neutral background, sticker of '))
```
julien-nc, line 11:
```suggestion
```

julien-nc (approved):
Remove the leftover debug console log and let's go.

Issue #321, opened 2025-07-29
Followup on file actions, support more file types


julien-nc:
After #304, as mentioned in https://github.com/nextcloud/assistant/pull/312#discussion_r2237232809 we could support all text-like file types for the summarize and TTS actions.

PR #326, opened 2025-08-05
feat: add support for more file types to file actions

lukasdotcom:
Resolves: #321 
Also changes the audio file types to be anything in VALID_AUDIO_MIME_TYPES so there aren't 2 separate lists keeping track of this.


kyteinsky (approved):
:rocket:

julien-nc (approved):
:+1:

Issue #316, opened 2025-07-22
AI hint in Chat UI


janepie:
The Chat UI is missing the hint that content is AI generated and should be doublechecked. This could be added in a line above the input field:

<img width="1556" height="1070" alt="Image" src="https://github.com/user-attachments/assets/52e17dda-4208-4671-a12a-ce6dce704867" />

PR #329, opened 2025-08-07
feat: ai hint in chat

lukasdotcom:
image
Resolves: #316


lukasdotcom:
> Looks nice! I would maybe adjust the wording here as this hint is also shown when there is no output yet, and it's not possible to adjust anything. Maybe "Output shown here is generated by AI. Make sure to always double-check." Wdyt @marcoambrosini ?

I like your wording more and pushed a commit and updated the screenshot based on it.

src/components/ChattyLLM/ChattyLLMInputForm.vue
```diff
@@ -125,6 +125,9 @@
 				class="session-area__agency-confirmation"
 				@confirm="onAgencyAnswer(true)"
 				@reject="onAgencyAnswer(false)" />
+			<p class="session-area__disclaimer">
+				Output shown here is generated by AI. Make sure to always double-check and adjust.
```
julien-nc, line 129: This should be translated :grin:

PR #349, opened 2025-09-03
Use new assistant theme

### LINKED ISSUE #347: Update assistant theme

- [ ] Change the selected tab to primary button style
    Component is here: https://github.com/nextcloud-libraries/nextcloud-vue/pull/7398

- [ ] Use `--color-element-assistant-icon` for the Nextcloud assistant title. It's ok to assign the same background to the icon and text just like in the assistant secondary button.
    Varialbes: https://github.com/nextcloud/server/issues/54665

Image

Lib version shipping the change is 8.30.0
https://github.com/nextcloud-libraries/nextcloud-vue/releases/tag/v8.30.0
https://www.npmjs.com/package/@nextcloud/vue/v/8.30.0

julien-nc:
closes #347

* [x] Adjust header menu icon: use NcHeaderButton (that comes with a transparency gradient)
image

* [x] Adjust the main form title (icon + text color)
* [x] Change selected task type category style

image
image

If we want to keep supporting NC 30 and 31, we need to be careful with this custom style that is not included in NC < 32. There is now a `use-modern-style` initial state injected to decide if we use the new css vars or not.

@marcoambrosini We can't use `NcAssistantButton` because
* NcActions is not flexible enough
* Even if we could use it to replace our `NcButton` in some places, this component does not allow to use a custom icon so we're stuck with the assistant one but we want the category one

So I went with custom styling using the css variables.
I tried to replicate the gradient in the NcAssistantButton component but this is done by using a gradient background in an element wrapping the NcButton. Here we directly use the button, no wrapping element. I couldn't find a way to have this gradient border so I removed the border which IMHO looks better than the border gradient + a different button gradient.
This is how NcAssistantButton look like with the 2 gradients:
image

### DIFF
--- a/lib/Listener/BeforeTemplateRenderedListener.php
+++ b/lib/Listener/BeforeTemplateRenderedListener.php
@@ -61,6 +61,9 @@ class BeforeTemplateRenderedListener implements IEventListener {
 		$assistantEnabled = $adminAssistantEnabled && $userAssistantEnabled;
 		$this->initialStateService->provideInitialState('assistant-enabled', $assistantEnabled);
 		if ($assistantEnabled) {
+			$useModernStyle = version_compare($this->config->getSystemValueString('version', '0.0.0'), '32.0.0', '>=');
+			$this->initialStateService->provideInitialState('use-modern-style', $useModernStyle);
+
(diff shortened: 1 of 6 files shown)


julien-nc:
With classic border and color being `#40519A`:
image

marcelklehr:
:tada:

PR #356, opened 2025-09-08
Fix stop polling when needed

### LINKED ISSUE #355: Stop polling a task when needed

2.7.1

33

Firefox

### Describe the Bug

When closing the assistant, if a task status was being polled, it continues because we don't cancel the polling loop. We should make it safer using the `destroyed` method of the component that manages this polling loop.

In some other cases like when selecting another task that is not scheduled/running, the polling loop is stopped correctly.

### Expected Behavior

There should never be any polling loop remaining after the assistant has been closed.

### To Reproduce

Schedule a task (or send a chat message) and close the assistant while the status is being polled. The polling loop is still active.

julien-nc:
closes #355

* Fix stop polling when closing the assistant while polling a classic task
* Fix stop polling when closing the assistant while polling a message generation in the chat UI (wrong component method name, `onDestroy()` does not exist anymore in vue3. We now use `beforeUnmount()`.
* Bonus: Get rid of the eslint warning about the `axios.isCancel` import

### DIFF
--- a/src/assistant.js
+++ b/src/assistant.js
@@ -313,6 +313,7 @@ export async function pollTask(taskId, obj, callback = updateTask) {
 				console.debug('[assistant] poll request failed', error)
 				if (error.status === 404) {
 					clearInterval(window.assistantPollTimerId)
+					window.assistantPollTimerId = null
 					reject(new Error('task-not-found'))
 					return
 				}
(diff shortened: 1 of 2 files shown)


kyteinsky (approved):
:rocket:

PR #360, opened 2025-09-15
Add pulse animation to assistant icon

### LINKED ISSUE #359: Hover animation of assistant button
> All done but the hover animation of the assistant button in the topbar 

 _Originally posted by @marcoambrosini in [#298](https://github.com/nextcloud/assistant/issues/298#issuecomment-3273586414)_

New suggestion from @jancborchardt: animate the size of the stars of the assistant icon by making them "pulse" one by one by one from top to bottom

marcoambrosini:
fix #359


marcoambrosini:
Removed review request as icon size shrunk when I made the viewport bigger, looking for workaround

jancborchardt:
@marcoambrosini I'd say generally we should treat hover and focus as close as possible, so yes. But it's not a blocker.

marcoambrosini:
Gave it a quick try and this is a bit complex. I will reference in new issue

marcoambrosini:
That would result in this animation being in many other places right? I think it would be better to avoid that for now

PR #395, opened 2025-11-11
fix: reinitialize inputs on New Task button click

### LINKED ISSUE #383: Clicking on "New Task" in ContextChat view permanently disables submit button

cloud.nextcloud.com

32.0.1-rc2

143.0.4 (64-bit) 

### Describe the Bug

Clicking on "New Task" in ContextChat view permanently disables submit button

### Expected Behavior

The submit button should be enabled as soon as I type something into the input.

### To Reproduce

1. Open Assistant
2. Open ContextChat Task view
3. Click on New Task
4. Type into input
5. Click on submit button

Result: Button not clickable

rroma:
- This fix adds reinitialization of context chat input form on the New Task button click

Fixes #383

### DIFF
--- a/src/components/ContextChat/ContextChatInputForm.vue
+++ b/src/components/ContextChat/ContextChatInputForm.vue
@@ -260,12 +260,10 @@ export default {
 		taskType(newValue) {
 			this.onTaskTypeChange()
 		},
-		inputs: {
-			handler(newValue) {
-				if (Object.keys(newValue).length === 0) {
-					this.onInputsReset()
-				}
-			},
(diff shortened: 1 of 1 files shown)


julien-nc:
Nice, it fixes the issue.
Any reason you defined the handler? Since you are not adding a "deep" watcher you can make the watcher just like the other ones, like the `taskType(newValue) {` one for example.

rroma:
Makes sense, thank you. I’ve just removed the handler.

janepie (approved):

PR #406, opened 2025-11-26
feat: Add system tags to AI generated files

marcelklehr:


lib/Service/SystemTagService.php
```diff
@@ -0,0 +1,43 @@
...
+		}
+	}
+
+	/**
+	 * @param string $fileId
+	 * @return void
+	 * @throws TagNotFoundException
+	 */
+	public function assignAiTagToFile(string $fileId) {
+		$this->systemTagObjectMapper->assignTags($fileId, 'files', $this->getAiTag()->getId());
```
kyteinsky: perhaps the tag can be created in a migration so we're sure it exists and then store the tag's ID in a config. It might be a little better to not create `ISystemTag` every time or is it not that impactful?

kyteinsky: > perhaps the tag can be created in a migration so we're sure it exists

can they be deleted by users? If yes, then ignore this review.

PR #419, opened 2025-12-09
Fix: no last task type

janepie:


src/assistant.js
```diff
@@ -409,7 +409,14 @@ async function getLastSelectedTaskType() {
 		},
 	}
 	const url = generateUrl('/apps/assistant/config')
-	return axios.get(url, req)
+	return axios.get(url, req).data.catch(error => {
```
marcelklehr: This won't work, I think. `axios.get` returns a promise
```suggestion
	return axios.get(url, req).catch(error => {
```

src/assistant.js
```diff
@@ -409,7 +409,14 @@ async function getLastSelectedTaskType() {
...
+		if (error.response?.status === 404) {
+			console.debug(t('assistant', 'No last task type available, falling back to default'))
+			return 'chatty-llm'
```
marcelklehr:
```suggestion
			return {data: 'chatty-llm'}
```

janepie:
Thanks, fixed

marcelklehr (approved):
Looks good, didn't test

PR #422, opened 2025-12-10
feat: Implement memories

marcelklehr:
Depends on

 - https://github.com/nextcloud/integration_openai/pull/310
 - https://github.com/nextcloud/llm2/pull/158
 - https://github.com/nextcloud/context_agent/pull/93

But is backward compatible

 - [x] Toggle in the chat
 - [x] Memories overview in the personal settings
 - [x] Design review
 - [x] Turn button into toggle and make toggle disappear in 3dot menu on small screens
 - [x] Don't load full sessions in setting page (because conversation_token is quite large)

### DIFF
--- a/lib/Settings/Personal.php
+++ b/lib/Settings/Personal.php
@@ -98,7 +98,7 @@ class Personal implements ISettings {
 
 		$rememberedSessions = $this->sessionMapper->getRememberedUserSessions($this->userId);
 		$rememberedSessionsShort = [];
-		foreach($rememberedSessions as $session) {
+		foreach ($rememberedSessions as $session) {
 			$rememberedSessionsShort[] = [
 				'id' => $session->getId(),
 				'title' => $session->getTitle(),


appinfo/routes.php
```diff
@@ -38,6 +38,7 @@
 
 		['name' => 'chattyLLM#newSession', 'url' => '/chat/new_session', 'verb' => 'PUT'],
 		['name' => 'chattyLLM#updateSessionTitle', 'url' => '/chat/update_session', 'verb' => 'PATCH'],
+		['name' => 'chattyLLM#updateSessionIsRemembered', 'url' => '/chat/update_is_remembered', 'verb' => 'PATCH'],
```
julien-nc: This is a good moment to make a generic endpoint to update a session like `PATCH /chat/session/{sessionId}` which would be used to update the session title and to update is_remembered.

julien-nc (changes requested):
So we need something in the frontend to toggle is_remembered for a session, right?

julien-nc (approved):
I hope you don't mind, I pushed a few commits to fix a typo and adjust the "update chat session" controller method.

marcelklehr:
Can you push your code?

marcelklehr:
Screenshot From 2025-12-15 09-18-42
Screenshot From 2025-12-15 09-18-34

marcoambrosini:
> Feels like no one will ever see or use this then

Copying from internal chat on the topic:
From my PoV it's ok. We also have not tested the usefulness of of this in the real world so I would prefer a conservative approach of having it behind a menu until we know it's really good and useful. We can always pop it out of the action menu as an unlabelled (or even labelled) button if we get lots of requests or when the feature is more mature.

janepie:
> We also have not tested the usefulness of of this in the real world so I would prefer a conservative approach of having it behind a menu until we know it's really good and useful. We can always pop it out of the action menu as an unlabelled (or even labelled) button if we get lots of requests or when the feature is more mature.

@marcoambrosini  We develop it for people to use it and we want it to be seen and used. Do you have any proposal for a location that is less "disruptive" and still visible?

I changed the icon to the memory one (like that!), as it doesn't have a filled version I adapted the text to change to Remembered when clicked, wdyt?

Screenshot_20251223_105549


image

julien-nc:
Maybe a tradeoff would be to make the button "icon-only" (with the label becoming the html title) and add a NcActionCheclbox item in the dropdown menu.

janepie:
New proposal by @nimishavijay : make it a toggle instead of a primary button so it's less "big block"

image

lib/Controller/ChattyLLMController.php
```diff
@@ -190,6 +192,49 @@ public function updateSessionTitle(int $sessionId, string $title): JSONResponse
...
+	 * @param string|null $title The new chat session title
+	 * @param bool|null $is_remembered The new is_remembered status: Whether to remember the insights from this chat session across all chat session
+	 * @return JSONResponse<Http::STATUS_OK, list{}, array{}>|JSONResponse<Http::STATUS_INTERNAL_SERVER_ERROR|Http::STATUS_NOT_FOUND, array{error: string}, array{}>
+	 *
+	 * 200: The title has been updated successfully
+	 * 404: The session was not found
+	 */
+	#[NoAdminRequired]
+	#[OpenAPI(scope: OpenAPI::SCOPE_DEFAULT, tags: ['chat_api'])]
+	public function updateChatSession(int $sessionId, ?string $title = null, ?bool $is_remembered = null): JSONResponse {
```
kyteinsky: maybe it would be nicer to indicate the metadata part

```suggestion
	 * Update session's metadata
	 *
	 * @param integer $sessionId The chat session ID
	 * @param string|null $title The new chat session title
	 * @param bool|null $is_remembered The new is_remembered status: Whether to remember the insights from this chat session across all chat session
	 * @return JSONResponse<Http::STATUS_OK, list{}, array{}>|JSONResponse<Http::STATUS_INTERNAL_SERVER_ERROR|Http::STATUS_NOT_FOUND, array{error: string}, array{}>
	 *
	 * 200: Metaadata of the session has been updated successfully
	 * 404: The session was not found
	 */
	#[NoAdminRequired]
	#[OpenAPI(scope: OpenAPI::SCOPE_DEFAULT, tags: ['chat_api'])]
	public function updateChatSessionMetadata(int $sessionId, ?string $title = null, ?bool $is_remembered = null): JSONResponse {
```

lib/Service/SessionSummaryService.php
```diff
@@ -0,0 +1,108 @@
...
+
+				$prompt = "Summarize insights about the user's circumstances, preferences and choices from the following conversation. Be as concise as possible. Do not add an introductory sentence or any other remarks.\n\n";
+
+				foreach ($messages as $message) {
+					$prompt .= $message->getRole() . ': ' . $message->getContent() . "\n\n";
+				}
+
+				$task = new Task(TextToText::ID, [
+					'input' => $prompt,
+				], 'assistant', $session->getUserId());
```
kyteinsky: the summary task type might be more suited to this since it also does chunking internally now

lib/Controller/ChattyLLMController.php
```diff
@@ -990,6 +1036,9 @@ private function scheduleLLMChatTask(
 			'system_prompt' => $systemPrompt,
 			'history' => $history,
 		];
+		if (isset($this->taskProcessingManager->getAvailableTaskTypes()[TextToTextChat::ID]['optionalInputShape']['memories'])) {
+			$input['memories'] = $this->sessionSummaryService->getUserSessionSummaries($this->userId);
+		}
```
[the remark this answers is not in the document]
marcelklehr: I'm not particular about that. We might save tokens and time that way, on the other hand the memories could aid in finding a good title. 🤷

lib/Service/SessionSummaryService.php
marcelklehr: The summary task type is to generic, IMHO. Also, we limit the amount of messages that go into the prompt here, so should not go beyond the context horizon

lib/Controller/ChattyLLMController.php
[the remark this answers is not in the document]
kyteinsky: yeah depends I suppose, it might remember something general about the user or may be unrelated from some other chat session, let's not change it then.

marcoambrosini:
> We develop it for people to use it and we want it to be seen and used. Do you have any proposal for a location that is less "disruptive" and still visible?

I'm ok with @nimishavijay suggestion as long as this action gets pushed into the menu on narrow screens.

PR #424, opened 2025-12-11
feat: Add buttons to suggest agency tools

marcelklehr:
[Kooha-2025-12-11-15-02-25.webm](https://github.com/user-attachments/assets/46246e98-6483-4187-9ca0-0b2031f0608b)

These buttons would only be present if context_agent is enabled.


marcoambrosini (commented):
I'd use tertiary buttons here and a flex layout for the container, wrapping them only when there's no available space.

marcelklehr:
> I'd use tertiary buttons here and a flex layout for the container, wrapping them only when there's no available space.

@marcoambrosini Good point! Thanks! What do you think about the general idea?

marcelklehr:
> What’s the weather tomorrow?

Agreed. Context agent doesn't offer this, yet, though :S

marcelklehr:
>  "What’s the weather today?", automatically taking your city from the profile

Yep, we can do that :)

PR #436, opened 2026-01-06
Make the notecards about missing features less alarming

julien-nc:
Make them info instead of warning.


janepie (approved):

julien-nc:
We could make a difference between "the task type has been disabled" and "there is no provider for the task type".

PR #437, opened 2026-01-08
Prepare 3.0.0

julien-nc:
### Added

- Add buttons to suggest agency tools @marcelklehr [#424](https://github.com/nextcloud/assistant/pull/424)

### Fixed

- Adjust to the breaking changes in nc/files @julien-nc [#428](https://github.com/nextcloud/assistant/pull/428)


marcelklehr:
Would be good to mention the new min version in the CHANGELOG. Also, I guess we can then start cleaning up backwards compatibility patches. Shall we create an issue for this?

PR #450, opened 2026-01-28
Expose task categories in getAvailableTaskTypes

lukasdotcom:
Moves the category determining code onto the server so that other clients can see the relevant category


marcelklehr (approved):
🎉

PR #451, opened 2026-02-03
Fix(ChattyLLM): Fix title generation to work with languages other than english

marcelklehr:
tested with Llama 3.1 8B and Mistral Small 24B

The current prompt would fail with Mistral Small 24B (tested on c.nc.c), ie. it would output swedish titles if sweden was mentioned in the conversation.

### DIFF
--- a/lib/Controller/ChattyLLMController.php
+++ b/lib/Controller/ChattyLLMController.php
@@ -895,10 +895,7 @@ class ChattyLLMController extends OCSController {
 				return new JSONResponse(['error' => $e->getMessage()], Http::STATUS_BAD_REQUEST);
 			}
 			return new JSONResponse(['taskId' => $taskId]);
-		} catch (\OCP\DB\Exception $e) {
-			$this->logger->warning('Failed to generate a title for the chat session', ['exception' => $e]);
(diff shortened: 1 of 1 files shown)


lib/Controller/ChattyLLMController.php
```diff
@@ -880,30 +880,27 @@ public function generateTitle(int $sessionId): JSONResponse {
...
 			} catch (\Exception $e) {
 				return new JSONResponse(['error' => $e->getMessage()], Http::STATUS_BAD_REQUEST);
 			}
 			return new JSONResponse(['taskId' => $taskId]);
 		} catch (\OCP\DB\Exception $e) {
 			$this->logger->warning('Failed to generate a title for the chat session', ['exception' => $e]);
 			return new JSONResponse(['error' => $this->l10n->t('Failed to generate a title for the chat session')], Http::STATUS_INTERNAL_SERVER_ERROR);
+		} catch (\JsonException $e) {
+			$this->logger->warning('Failed to generate a title for the chat session', ['exception' => $e]);
+			return new JSONResponse(['error' => $this->l10n->t('Failed to generate a title for the chat session')], Http::STATUS_INTERNAL_SERVER_ERROR);
```
julien-nc: Nitpick: You could use `|` and merge the 2 catch blocks.

marcelklehr: Done

Issue #408, opened 2025-11-28
Show provider information in the UI


julien-nc:
We could let the users know more about which provider is going to process their tasks. We could also potentially display the AI rating of the provider.

For that we need to:

* Include some more provider-related information in the task type list for the Assistant (and clients) to display it in their UI
* Maybe extend the provider concept so it includes a description and a rating
* Figure out a way to compute the rating in integration_openai depending on the service that is used. We could have a hardcoded list of service URLs for which we know the rating. If an unknown URL is set, we could let the admin set the rating in integration_openai's admin settings.

cc @marcelklehr @DaphneMuller

DaphneMuller:
closing as completed by https://github.com/nextcloud/assistant/pull/483

PR #483, opened 2026-03-02
feat: Show provider information in the UI (#408)

zainabnaseer1164:
Shows which AI provider will process each task type in the Assistant UI, so users know which provider handles their request.

### DIFF
--- a/src/components/AssistantTextProcessingForm.vue
+++ b/src/components/AssistantTextProcessingForm.vue
@@ -55,7 +55,7 @@
 							<EditableTextField :initial-text="selectedTaskType.name ?? ''" />
 							<span v-if="selectedTaskType.preferredProviderName"
 								class="session-area__top-bar__provider">
-								{{ t('assistant', 'Provider:') }} {{ selectedTaskType.preferredProviderName }}
+								{{ t('assistant', 'Provider: {name}', { name: selectedTaskType.preferredProviderName }) }}
 							</span>
 						</div>
 					</div>


julien-nc (changes requested):
Hi! Thanks for the PR!

* Could you explain a bit more what you did?
* Could you provide a screenshot (useful to get an idea of what should be checked/tested)?
* Could you sign your commits? You can set a global GPG key for git (see the `GPG Introduction` section of https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work) and then follow the instructions of the failing DCO GitHub action)
* Could you generate the OpenAPI specs? (the openapi composer script does that)
* Could you move the new provider info in a place where the layout is not affected? Like next to the task type title? Right now it takes a lot of vertical space.
* There's something wrong with some characters, can you check that? 

Image

Feel free to ask questions if something I said was not clear.

zainabnaseer1164:
> Thank you for the adjustments!
> 
> > Openapi.json already includes preferredProviderName in this PR.
> 
> But the OpenAPI CI action is still failing because the specs were not generated again so they are out of date.
> 
> > UI shows it next to the task type title and in the dropdown as “TaskName · ProviderName”.
> 
> I think adding the provider name in the dropdown is a bit too much. It is enough to have it in the form title. Wdyt?
> 
> > I’ll fix display so apostrophes (e.g. “OpenAI’s”) show correctly instead of wrong one.
> 
> I can see it's displayed correctly now, did you fix it already?
> 
> Image
> > So I will do it once I get approval from your side.
> 
> Sure if there is still a visual issue, you have my approval to fix it.

@julien-nc above comments have been tackled along with passing all pending GitHub test.

Issue #495, opened 2026-03-24
User Avatar is always white


marcelklehr:
### Which version of assistant are you using?

latest

### Which version of Nextcloud are you using?

33.0.1-rc

### Which browser are you using? In case you are using the phone App, specify the Android or iOS version and device please.

latest Firefox

### Describe the Bug

<img width="505" height="200" alt="Image" src="https://github.com/user-attachments/assets/6319a1a8-76dc-4e80-ba26-b7a6c99ec6b8" />

### Expected Behavior

Assistant should use my avatar like other places in nextcloud

### To Reproduce

1. Open assistant chat
2. Write a message
3. Avatar is white

marcelklehr:
Supposedly fixed by #496 

but on c.nc.c I can still observe this with v3.4.0:

<img width="372" height="135" alt="Image" src="https://github.com/user-attachments/assets/b3f54530-4178-4fe9-922d-32bd685bd438" />

PR #496, opened 2026-03-24
Fix chat message avatar

julien-nc:
The `#icon` template of the message avatar can be empty. It seems like with previous nc/vue releases, this would fall back to `user` or `displayName`. This is not the case anymore so we always use this template.

Solution: only declare the template when we need it. Otherwise keep using the props.


marcelklehr (approved):

marcelklehr:
Seems like this fix didn't work #495

Issue #284, opened 2025-06-23
Draggable and resizable micro-chat


marcoambrosini:
## Problem: 
The chat with assistant feature can be disruptive for certain workflows because the assistant completely blocks the current app UI 

## Describe the solution you'd like

It would be nice to be able to access the chat with AI functionality from an **overlaying** and **non-blocking** draggable window. That way, Alice can continue her work in  after asking a question to the assistant in the chat. 
This feature could be triggered by a keyboard shortcut, and/or by having a minimize button in the assistant view. This micro-chat view could in turn have a maximise button to invoke the full size assistant.

Quick, non legally binding mock up only for illustrative purposes: 
<img width="1248" alt="Image" src="https://github.com/user-attachments/assets/a0cb9936-1998-4f26-b0c1-8c438b9007e0" />

## Competitive analisys
This is a feature that several competitors already offer. They all have their main big chat UI, but they also offer the micro chat, which I assume ends up being the main entry point for many users.

<img width="189" alt="Image" src="https://github.com/user-attachments/assets/2ae0b924-df88-4f76-9edf-9d3421e9ddc5" />
<img width="772" alt="Image" src="https://github.com/user-attachments/assets/1c63138f-2d03-4c64-b99e-4f3cd36bda8f" />
<img width="336" alt="Image" src="https://github.com/user-attachments/assets/e64509c2-53e7-4b70-900f-c69d91112f55" />

@julien-nc and @DaphneMuller since we discussed this a moment ago

PR #507, opened 2026-04-09
Make the modal non modal

julien-nc:
The goal of this change is to keep the ability to interact with the current page when the assistant is displayed on top.
Instead of using NcModal, we switch to PimeVue's dialog that can be moved and can be non-modal.

* [x] use PrimeVue's dialog
* [x] make it resizable
* [x] adjust the style everywhere so the assistant UI behaves the same and the standalone page is not broken
* [x] merge the dialog header with the assistant title (keep the ability to move the dialog)
* [x] use the grab and grabbing cursor when hovering and dragging the dialog header
* [ ] force the fullscreen mode on narrow screens (mobile)
* [x] set a NC-like style to the fullscreen button

[Kooha-2026-04-08-18-49-41.webm](https://github.com/user-attachments/assets/d38c3f46-2c5a-418c-9539-3b9ba7e9d2a7)

@marcoambrosini Wdyt?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,9 +23,11 @@
         "@nextcloud/moment": "^1.3.1",
         "@nextcloud/router": "^3.0.0",
         "@nextcloud/vue": "^9.0.0-rc.5",
+        "@primeuix/themes": "^2.0.3",
         "extendable-media-recorder": "^9.2.11",
         "extendable-media-recorder-wav-encoder": "^7.0.129",
         "moment": "^2.30.1",
+        "primevue": "^4.5.4",
         "v-click-outside": "^3.2.0",
         "vue": "^3.5.16",
         "vue-material-design-icons": "^5.1.2"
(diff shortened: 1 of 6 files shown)


marcoambrosini (commented):
Looks nice @julien-nc, but since this is a pattern that we do not have yet in Nextcloud, I would loop in @nextcloud/designers for their thoughts.

From my POV there is value in non blocking the content with the assistant dialog, but:

* it shouldn't fall on the user to decide where to place the content
* we should probably make room for it in the UI instead of having an overlay.

To me the perfect container for this is a sidebar, which also happens to be the solution that most other products use.

Thoughts?

julien-nc:
Yeah this is just a PoC.

The problem is the sidebar is too narrow and it would require to redesign the chat UI to make it fit such narrow space. If the sidebar width could be adjusted then maybe.

My thoughts:
* In any case, we need at least 650px of width to display the chat UI and keep it usable
* We need a way to still show the modal or something similar if the user does not care about loosing the focus on the rest of the page

marcoambrosini:
> we need at least 650px

Why do you think so? In talk we fit the chat into the sidebar during calls and it works reasonably well. That said, we could still think about introducing custom widths to the component for both use cases. 

> We need a way to still show the modal or something similar if the user does not care about loosing the focus on the rest of the page

Agreed! Something like the expand button you have here would work well I think :)

julien-nc:
> Why do you think so? In talk we fit the chat into the sidebar during calls

In Talk the conversation selector is in the left/navigation sidebar. For the assistant we have to fit the whole UI in the sidebar which includes the conversation selector. So either we:
* redesign the entire assistant UI so it works when it's very narrow
* OR we give it enough width so we can keep the current design

julien-nc:
Btw @marcoambrosini you were the author of the initial request: #284 :grin: .
We think this change is fine, it looks a lot like before, the main difference is that the users can keep working on the current page while using the chat UI.

DaphneMuller:
@marcoambrosini to add some context, your issue https://github.com/nextcloud/assistant/issues/284 has been on our to do list for quite a long time but we struggled finding a pragmatic way to implement it. After quite some brainstorms the team came up with this solution which in my opinion works very nicely and meets the requirements outlined in the issue. I think it would be a cool component to have in other apps of Nextcloud as well, for example I can see benefit in having a floating Notes app or a floating Tasks app.

The user has a good default with a pretty large screen that looks almost identical to what we have today. So I don't believe we add an additional burden on the user.

Is this really a blocker for you?

nimishavijay:
@marcoambrosini I am probably missing a lot of previous context and some of these points may be existing/closed issues but just my 2c: I'd agree that a sidebar is more ideal, but having any non-blocking view is valuable. some suggestions to make this more easy to use:
- automatically close the left history list if it's not the fullscreen view
- allow minimizing the chat into a little messenger-like floating thingy in the corner (similar to a draft in Mail). This would also work great as a shortcut for scheduled tasks (eg: summarize a text file from its action menu currently results in a toast saying "Summarization AI task submitted successfully. You will be notified when it is ready. It can also be checked in the Assistant in the "Work with text -> Summarize" menu", it could instead just show up in the little corner element) 
- displaying the full screen non-modal whenever the top-right icon is clicked and the smaller draggable non-modal whenever assistant is invoked from anywhere other than the top right icon. (so creating a distinction of "when I click on the assistant app it is a full blocking thing, when i enter the assistant from anywhere else it is a quick floating thing")
- visually it feels a bit flat IMO, could use stronger shadows in the final version

marcoambrosini:
@julien-nc @DaphneMuller #284 is about a tiny chat UI, not the whole assistant interface. This pr to me looked to me like a different project. 

Thanks @nimishavijay for your input, Indeed we would need to hide things to improve this. I like the idea of minimizing too. 

> Is this really a blocker for you?

Didn't say there were any blockers :) 

What do you think about reducing the functionality available here and making this window smaller?

DaphneMuller:
In the future we can indeed look into minimizing the history window when the screen is not large enough. I think that's a good idea. We need to keep the features outside of the chatUI because we have a userbase of which not all people are great prompters. I will ask Julien to merge this PR and then we can open new issues for things we want adjusted in the future.
@julien-nc can you create an issue for minimizing the history window?

janepie:
Looks nice! Two issues I found with changing the window size with an open dialog:
- if I changed the dialog size and moved it to the size, then adapt the window size, it doesn't adjust its position and gets cut off at the side
chrome-capture-2026-4-17

- If I close the History sidebar, make the window smaller and then bigger again, the sidebar opens again

Additionally it would be nice if the position and size is remembered when I close and open the dialog, but I think that's a new issue :)

julien-nc:
@janepie Thanks for the feedback.

* Gets out of the window after window resize: It sounds fine to me, it can be moved in again. I don't think we can avoid that.
* The sidebar opens when the window width grows: This is the behaviour of NcAppNavigation. Not sure we can prevent that. It's not so bad as it can be closed again.
* Remember the size and position: It's tough to restore size+position when we don't know if the current window size is the same as when the size+position was saved.  I would save that for later.

jancborchardt (changes requested):
I think it’s fine to do it like this, it’s an improvement over the current state. If you don’t know you can resize and move, it’s basically the same. If you do know it, it’s an enhancement.

Moving it to a sidebar like is done by VS Code or Firefox for example is something that has to be discussed way more, so should not be a blocker for this.

Just as discussed with @julien-nc things to adjust:
- [x] The height of the dialog should be a bit reduced as right now it uses almost all vertical space.
- [x] There is a jump in width initially before the content is loaded, which shouldn’t happen. The dialog should directly have it’s final size, and ideally not just with a spinner, but with skeleton screen for left navigation and content.
- [x] As @nimishavijay mentioned, the shadow needs to be proper enough so that some amount of focus is still given even though we can’t darken the background anymore.

PR #514, opened 2026-04-27
fix(l10n): Use "teams" instead of the old "circles"

marcelklehr:


lib/Service/AssistantService.php
```diff
@@ -94,21 +94,21 @@ public function __construct(
 	) {
 		$this->informationSources = [
 			'add_card' => $this->l10n->t('Nextcloud Deck'),
-			'add_member_to_circle' => $this->l10n->t('Add member to circle'),
```
rakekniven, line 97: @marcelklehr Are you sure about all changes for given file?
You replaced multiple miscallenous strings with "Nextcloud Teams"

marcelklehr: Yes, it's the cleaner way to reference the tool, as we show these as "sources". "Add member to circle" is not really semantically a source.

rakekniven (approved):

rakekniven:
@marcelklehr Backports needed to get rid of the "circle" strings at Tx?

marcelklehr:
Indeed, we have a stable32 branch

PR #523, opened 2026-05-05
Prevent opening the assistant dialog multiple times

julien-nc:
Now that the dialog is not modal anymore, it is possible to open it multiple times because we can interact with the UI elements that trigger that.

This fix prevents that from happening when it is opened from:
* the top right menu entry
* another app by calling `openAssistantForm`
* a notification

In a nutshell: We remember when it's opened and we remember when it is closed (the Vue instance is unmounted).

### DIFF
--- a/src/assistant.js
+++ b/src/assistant.js
@@ -11,8 +11,6 @@ import Aura from '@primeuix/themes/aura'
 
 window.assistantPollTimerId = null
 
-// TODO add param to lock on specific task type
-
 /**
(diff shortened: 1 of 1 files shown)


kyteinsky (approved):
just a suggestion: maybe the button can be disabled/used as a toggle in addition to this change
the toggle part maybe only works if the visibility is toggled, isn't great when the div is unmunted I suppose.

marcelklehr (approved):

PR #524, opened 2026-05-06
Fix: title height

janepie:
Is fixed to the header height now and scrolls when the title doesn't fit

image

### DIFF
--- a/src/components/ChattyLLM/ChattyLLMInputForm.vue
+++ b/src/components/ChattyLLM/ChattyLLMInputForm.vue
@@ -1058,8 +1058,9 @@ export default {
 
 			&__title {
 				width: 100%;
-				height: var(--header-height);
+				height: calc(var(--default-clickable-area) + var(--default-grid-baseline) * 2);
 				overflow: scroll;
+				align-content: center;
 			}
 
 			&__remember {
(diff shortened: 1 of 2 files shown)


miaulalala (approved):

julien-nc:
### With this branch

image

### With the main branch

image

Weird... For short titles, main looks more correct.

janepie:
What do you mean? Your title is too short

janepie:
Or do you mean the height of the line?

julien-nc:
Yes height of the line with a short title.

julien-nc:
And with a long title it scrolls vertically. I think horizontally is better. Wdyt?
Or even Ellipsis and add an HTML title (with the session title as value).

janepie:
Better?

Screencast above

marcelklehr:
Ellipsis and add an HTML title is the way to go, imho

janepie:
Not sure what you mean by "add an HTML title"

marcelklehr:
Add an html title attribute with the full text, as it will be cut off by the ellipsis.

janepie:
Done!

julien-nc (approved):
👍

PR #539, opened 2026-05-13
feat(Assignments): Send notifications for assignment events

marcelklehr:


julien-nc (approved):
LGTM 👍 
Later we can intercept the click on the notification action and just open the assistant instead of following the link.

PR #540, opened 2026-05-13
fix(assistant): render markdown in TextInput output

n-iv:
## Summary

The assistant's "generate text" form currently displays its output through NcRichContenteditable, which renders text as-is. When the underlying LLM returns markdown (heading, lists, tables, bold, links), the user sees raw markdown syntax instead of formatted content.

This PR wraps the output field in NcRichText when isOutput is true, falling back to NcRichContenteditable for the input case. Headings, lists, bold, italic, code blocks and links are now rendered. Tables are not rendered due to the markdown-it configuration of NcRichText in @nextcloud/vue, which does not enable the table plugin by default. The copy still functions and keeps the markdown as expected.

## Screenshots

**Before**
Screenshot_20260513_102348

**After**
Screenshot_20260513_130308

### DIFF
--- a/src/components/fields/TextInput.vue
+++ b/src/components/fields/TextInput.vue
@@ -13,6 +13,7 @@
 			v-if="isOutput && hasValue && !isEditing"
 			ref="richText"
 			class="rendered-output output-wrapper"
+			:class="{ streaming: isOutput && streaming() }"
 			:title="t('assistant', 'Double-click to edit')"
 			:text="value ?? ''"
 			:use-markdown="true"
@@ -353,8 +354,14 @@ body[dir="rtl"] .output-buttons {
 	.shadowed.streaming .rich-contenteditable__input {
 		animation: pulse 2s infinite;
 	}
+	.output-wrapper.streaming {
+		animation: pulse 2s infinite;
+	}
 	@media (prefers-reduced-motion: reduce) {
 		.shadowed.streaming .rich-contenteditable__input {
+			animation: none;
+		}
+		.output-wrapper.streaming {
 			animation: none;
 		}
 	}
--- a/src/components/fields/TextInput.vue
+++ b/src/components/fields/TextInput.vue
@@ -9,7 +9,16 @@
 			<br v-if="limitLabel">
 			{{ limitLabel ?? '' }}
 		</label>
+		<NcRichText
+			v-if="isOutput && hasValue && !isEditing"
+			class="rendered-output output-wrapper"
+			:title="t('assistant', 'Double-click to edit')"
+			:text="value ?? ''"
+			:use-markdown="true"
+			:autolink="true"
+			@dblclick="enterEditMode" />
 		<NcRichContenteditable
+			v-else
 			:id="id"
 			ref="input"
 			:model-value="value ?? ''"
@@ -21,7 +30,8 @@
 			:placeholder="placeholder"
 			:title="title"
 			@submit="hasValue && $emit('submit', $event)"
-			@update:model-value="$emit('update:value', $event)" />
+			@update:model-value="$emit('update:value', $event)"
+			@blur="onEditableBlur" />
 		<div v-if="isOutput && hasValue"
 			class="output-buttons">
 			<NcButton v-if="!streaming() && canImproveOutput"
@@ -73,6 +83,7 @@ import ContentCopyIcon from 'vue-material-design-icons/ContentCopy.vue'
 import NcButton from '@nextcloud/vue/components/NcButton'
 import NcRichContenteditable from '@nextcloud/vue/components/NcRichContenteditable'
 import { NcLoadingIcon } from '@nextcloud/vue'
+import { NcRichText } from '@nextcloud/vue/components/NcRichText'
 
 import isMobile from '../../mixins/isMobile.js'
 
@@ -98,6 +109,7 @@ export default {
 
 	components: {
 		NcRichContenteditable,
+		NcRichText,
 		NcButton,
 		NcLoadingIcon,
 		FileDocumentOutlineIcon,
@@ -156,6 +168,7 @@ export default {
 	data() {
 		return {
 			copied: false,
+			isEditing: false,
 			maxLength: MAX_TEXT_INPUT_LENGTH,
 		}
 	},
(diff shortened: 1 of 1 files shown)


janepie:
Thank you for this PR! The problem with this is that is makes the output uneditable for the user, and we assume that the LLM output is not perfect and has to be adapted. Can you find a way to keep that function, and also keep the "easy copy" button?

n-iv:
Hi @janepie, the easy copy button is still present and can be used. I added a very short video to show it.
[Screencast_20260518_101227.webm](https://github.com/user-attachments/assets/03254e06-9256-4a28-a8b4-7799b5974b91)

janepie:
Ah thanks for showing me! It's a bit unfortunate that the button and the AI-generation note disappear into the scrollable area, we should keep them visible and make the result scrollable instead.

So points to adapt would be

- [ ] keep an easy way to edit the result (for example change into edit mode with a doubleclick on the area)

- [ ] always keep copy button and AI note in sight and make the result scrollable instead

n-iv:
I pushed a new commit addressing both points:

- Double-clicking the rendered output now switches it to an editable contenteditable, which returns to rendered mode on blur. A `title` attribute hints at the affordance.
- The rendered output (and the editable variant in output mode) is now wrapped in a bounded scrollable container, so the result scrolls inside the component instead of pushing the copy button and the AI-generation note out of sight.
Screenshot_20260520_121515
Screenshot_20260520_121648
Screenshot_20260520_121825
Screenshot_20260520_121838

src/components/fields/TextInput.vue
```diff
@@ -9,7 +9,19 @@
 			<br v-if="limitLabel">
 			{{ limitLabel ?? '' }}
 		</label>
+		<div
```
janepie: Why wrapping it in a div? Should work with directly using NcRichText

src/components/fields/TextInput.vue
```diff
@@ -215,6 +264,7 @@ body[dir="rtl"] .choose-file-button {
 	.copy-button,
 	.choose-file-button {
 		position: absolute !important;
+		z-index: 10;
```
janepie: What's this for?

janepie (commented):
Looks good! 
I'd prefer to have the double-click hint somewhere more visible but I'm not sure yet where to put it, so let's do that in a follow-up. 

A few small remarks

src/components/fields/TextInput.vue
julien-nc: I tried removing it and couldn't see any difference. Is it fixing something in a very special case?

src/components/fields/TextInput.vue
```diff
@@ -226,6 +276,18 @@ body[dir="rtl"] .choose-file-button {
 		right: 4px;
 	}
 
+	.output-wrapper {
+		display: block !important;
+		box-sizing: border-box !important;
+		border: 2px solid var(--color-primary-element) !important;
+		border-radius: var(--border-radius) !important;
```
julien-nc: This should be `--border-radius-large` to be consistent with the NcRichContenteditable border radius.

src/components/fields/TextInput.vue
```diff
@@ -234,6 +296,8 @@ body[dir="rtl"] .choose-file-button {
 	.shadowed .rich-contenteditable__input {
 		border: 2px solid var(--color-primary-element) !important;
 		padding-bottom: 38px !important;
+		max-height: 35vh !important;
+		overflow-y: auto !important;
```
julien-nc: `.rich-contenteditable__input` already has `overflow-y: auto` set by the server style. This can be removed.

src/components/fields/TextInput.vue
julien-nc: If the wrapper can be easily removed, I agree with Jana it would be nice to remove it. But if it's way more convenient to have it because styling NcRichText directly is sketchy, then let's keep the wrapper.

src/components/fields/TextInput.vue
```diff
@@ -226,6 +276,18 @@ body[dir="rtl"] .choose-file-button {
...
+	.output-wrapper {
+		display: block !important;
+		box-sizing: border-box !important;
+		border: 2px solid var(--color-primary-element) !important;
+		border-radius: var(--border-radius) !important;
+		padding: 8px !important;
+		padding-bottom: 42px !important;
+		max-height: 35vh !important;
+		overflow-y: auto !important;
+		cursor: text;
```
julien-nc: This is not enough.
```suggestion
		.rendered-output, .rendered-output * {
			cursor: text;
		}
```

julien-nc (changes requested):
Nice change! I like the increase of the max-height.

To make this more discoverable, there could be an "edit" button in the top right corner when it's not editable. Wdyt @janepie and @n-iv ? It would require to slightly increase the min-height so this edit button does not overlap with the copy one.

I spotted a minor bug, when the field is empty, we can only type one character and it leaves the edit mode. A double click is necessary to continue editing. This does not happen when the field is not empty. I'd say it's fine if we don't find an easy fix for that.

src/components/fields/TextInput.vue
n-iv: Before:
Screenshot_20260525_211556

After:
Screenshot_20260527_235820

Done. Switched to --border-radius-large. Definitely prettier, thanks.

n-iv: The cursor is fixed elsewhere, btw.

src/components/fields/TextInput.vue
n-iv: Done. No difference as expected.

src/components/fields/TextInput.vue
n-iv: Before:
Screenshot_20260527_235820

After:
Screenshot_20260528_002159

That's where the cursor's fixed. No complication, rendering and double-click both still work.

src/components/fields/TextInput.vue
n-iv: Done

n-iv:
Screenshot_20260528_011643
Remove the block and the copy button ends up bottom-left and outside the output box instead of overlaying it. I'd keep it unless you see a different intent behind.

julien-nc:
Thanks for the adjustments.

For the style issue, I can't reproduce.

image

Are you sure you're not tweaking the CSS? Are you using the custom_css app?
The border-radius of the input field and of the task items in the left navigation sidebar is wrong on your side.

Or maybe your server-injected CSS is different. Which version of NC are you using?

julien-nc:
Also, you need to sign your commits. Check the [DCO check output](https://github.com/nextcloud/assistant/pull/540/checks?check_run_id=78186930878).

src/components/fields/TextInput.vue
n-iv: dropped it

n-iv:
Rebased on main and squashed into a single commit. Dropped the z-index as suggested => no visible difference, as @julien-nc noticed. And to answer that last question, yes we use a lot of css on our instances. Ready for another look.

julien-nc (commented):
* Rebased on main
* Adjusted the min height of NcRichText (we changed that in other places recently)
* Made sure NcRichText is scrolling to bottom while its content is streamed (recent change as well)
* Force pushed

One remaining question: What is the change in TaskTypeSelect.vue about? If it's unrelated, let's remove it.
Once we sort this out we can get this in!

n-iv:
Sorry for the extra change in TaskTypeSelect.vue, I removed it, it was a local workaround test.

julien-nc (approved):
Fixed a detail: Added the pulse animation when streaming.
Tested and approved

PR #544, opened 2026-05-18
Fix: throw UnknownActivityException on Notifier

janepie:
https://github.com/nextcloud/server/pull/59609

### DIFF
--- a/lib/Notification/Notifier.php
+++ b/lib/Notification/Notifier.php
@@ -7,8 +7,8 @@
 
 namespace OCA\Assistant\Notification;
 
-use InvalidArgumentException;
 use OCA\Assistant\AppInfo\Application;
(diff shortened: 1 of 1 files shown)


provokateurin (approved):

lib/Notification/Notifier.php
```diff
@@ -7,8 +7,8 @@
 
 namespace OCA\Assistant\Notification;
 
-use InvalidArgumentException;
 use OCA\Assistant\AppInfo\Application;
+use OCP\Activity\Exceptions\UnknownActivityException;
```
nickvergessen, line 11: In the notifier you need to throw `OCP\Notification\UnknownNotificationException`

the UnknownActivityException exception is only for activity parsing 🙈

janepie: aaah damn

PR #545, opened 2026-05-18
Streaming responses

julien-nc:
Depends on:
* https://github.com/nextcloud/server/pull/60167
* https://github.com/nextcloud/server/pull/60500
* https://github.com/nextcloud/server/pull/61197
* https://github.com/nextcloud/integration_openai/pull/366

This adds the ability to get and display intermediate (streamed) text responses in the chat UI and in the generic form.
We use `@nextcloud/notify_push` if the notify push service is configured and available. If not, we still get intermediate results with the (pre-existing) task status polling.

### Notify push

[Kooha-2026-05-18-13-27-53.webm](https://github.com/user-attachments/assets/46c41b28-b1d4-4dbb-9953-2aca55789630)

[Kooha-2026-05-19-17-04-29.webm](https://github.com/user-attachments/assets/543ae0bb-ec6c-4eb1-9872-af4ed8e58999)

### Polling (fallback)

[Kooha-2026-05-18-13-40-25.webm](https://github.com/user-attachments/assets/3c161833-e404-405a-a9e5-5b87f5d2a59c)

[Kooha-2026-05-19-17-07-20.webm](https://github.com/user-attachments/assets/706595ba-2550-4826-92d1-ab3089b42d76)

### DIFF
--- a/lib/Controller/ChattyLLMController.php
+++ b/lib/Controller/ChattyLLMController.php
@@ -601,7 +601,14 @@ class ChattyLLMController extends OCSController {
 		} elseif ($task->getstatus() === Task::STATUS_RUNNING || $task->getstatus() === Task::STATUS_SCHEDULED) {
 			$startTime = $task->getStartedAt() ?? time();
 			$slowPickup = ($task->getScheduledAt() + (60 * 5)) < $startTime;
-			return new JSONResponse(['task_status' => $task->getstatus(), 'slow_pickup' => $slowPickup], Http::STATUS_EXPECTATION_FAILED);
+			$responsePayload = [
+				'task_status' => $task->getstatus(),
+				'slow_pickup' => $slowPickup,
+			];
+			if ($task->getstatus() === Task::STATUS_RUNNING) {
+				$responsePayload['task_output'] = $task->getOutput();
+			}
+			return new JSONResponse($responsePayload, Http::STATUS_EXPECTATION_FAILED);
 		} elseif ($task->getstatus() === Task::STATUS_FAILED || $task->getstatus() === Task::STATUS_CANCELLED) {
 			return new JSONResponse(['error' => 'task_failed_or_canceled', 'task_status' => $task->getstatus()], Http::STATUS_BAD_REQUEST);
 		}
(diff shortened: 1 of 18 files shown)


DaphneMuller:
That is looking great and will definitely improve the experience of performance for users! 
Small nitpick and not critical, but it could be nice to indicate that the answer is still loading for the polling - work with text example.

julien-nc:
There is an indicator already, there is a loading icon next to the task type name at the top of the form (below the task type list):
image

I agree this is not enough. Nothing is definitive in this "WIP" PR.

How would you improve this?

I was thinking of imitating what's done in some other tools: Make the text field border pulse. Right now it is a thick, static, primary color border. It could pulse from border-color to primary color.

DaphneMuller:
I don't really know what the best solution is, maybe @marcoambrosini can help with that. Keep in mind that things are good regarding people who don't use animations for motion sickness. Pulsing the border color could be an idea

marcoambrosini (commented):
Nice addition!

I think a good solution for this it could be a blinking informative line of text below the generated output (e.g. “generating text”). The advantage of the text is that it could not blink if reduce motion option is toggled at OS level, but the information would still be there.

Here's an example of that in perplexity:

Screenshot 2026-05-19 at 14 56 51

I would not rely on the border of the input field for this because I think these input fields should not be input fields at all, only output fields :p 
IDK if this is something you’ve talked about already, but I’d much rather have users “save output to text file” if they want to edit the output. Or manually copy and paste to whatever app they're working in. I don't think anyone really "works with text" in this tiny textarea element.

julien-nc:
@marcoambrosini Thanks for the feedback.

In the chat UI, while a message is being generated, we already have a placeholder in the input field. This placeholder could blink/pulse while streaming the response.

In the generic form (all task types other than chat), we are not using a chat-like layout. I don't know where a "thinking" label would appear.

I already added a line in the form header (a label and 2 buttons). The "Generating results..." label could blink/pulse. Wdyt?

julien-nc:
@marcoambrosini I updated the 2nd and 4th screencasts (generate text). There is now a pulse animation around the output fields and the "copy" button's label is "Generating results..." while streaming.

marcoambrosini:
@julien-nc sounds good!

marcelklehr:
Added a commit that introduces pseudo streaming to make it more appealing visually and give the illusion of a continuous stream of text.

marcelklehr (approved):
🐘

PR #522, opened 2026-05-05
feat(Assignments)

marcelklehr:
Todo:

 - [x] Refactored ChattyLLMController to factor out business logic into a service
 - [x] Implemented Assignments backend logic (db, APIController, Service, BgJob)
 - [x] Implemented RRule checking
 - [ ] Implement front-end
     - [ ] Small header for assignment sessions to indicate what's going on
     - [ ] Maybe disable the chat input for assignment sessions?
     - [ ] Maybe a small icon for the chat session overview for assignment sessions
     - [ ] Query the session list regularly and for assignment sessions we need to check for new generations regularly
 - [ ] Implement better session titles for assignment sessions
 - [ ] Implement notification for mid-assignment approvals

Use the assignment tools in https://github.com/nextcloud/context_agent/pull/162 to schedule assignments from the UI, or use curl :)

PR #550, opened 2026-05-21
Implement scheduled tasks ui

lukasdotcom:
Resolves: #522
image
image


lib/Db/ChattyLLM/MessageMapper.php
```diff
@@ -80,17 +80,22 @@ public function getLastNonEmptyHumanMessage(int $sessionId): Message {
 	 * @param int $sessionId
 	 * @param int $cursor
 	 * @param int $limit
+	 * @param bool $hideUserMessages
 	 * @return list<Message>
 	 * @throws \OCP\DB\Exception
 	 */
-	public function getMessages(int $sessionId, int $cursor, int $limit): array {
+	public function getMessages(int $sessionId, int $cursor, int $limit, bool $hideUserMessages = false): array {
```
marcelklehr: Mh, I wonder if we can refine this param a bit. First off, it's formulated in a negative way, so if you *don't* want to see the user messages, you set it to `true`, which can be confusing. Better would be `showUserMessages` with a default to `true`. But Perhaps it would be even better as a `filterByRole` option, where you can pass all roles. What do you think?

src/components/ChattyLLM/ChattyLLMInputForm.vue
```diff
@@ -290,16 +340,20 @@ export default {
...
+					message: t('assistant', 'What\'s the weather today?'),
+				},
+				{
+					aria: t('assistant', 'Ask assistant, to create a share link for a file'),
+					message: t('assistant', 'Can you create a share link for me?'),
 				},
 				{
-					aria: t('assisant', 'Ask assistant, to create a share link for a file'),
-					message: t('assisant', 'Can you create a share link for me?'),
+					aria: t('assistant', 'Ask assistant, create a scheduled task to send me the weather every morning'),
```
marcelklehr: Nice! I'd readd the "to", though, no?

src/components/ChattyLLM/recurrenceRule.js
```diff
@@ -0,0 +1,236 @@
...
+				monthNames,
+				dayOfMonthList,
+			})
+		} else {
+			const ordinalNumber = getTranslatedOrdinalNumber(recurrenceRule.bySetPosition)
+			const byDaySet = getTranslatedByDaySet(recurrenceRule.byDay)
+
+			// Prevent empty text in those sections
+			if (ordinalNumber !== '' && byDaySet !== '') {
+				limitPart = t('assistant', 'in {monthNames} on the {ordinalNumber} {byDaySet}', {
```
marcelklehr, line 120: mmh, this is likely gonna break in other languages. For example in Portuguese the ordinal number changes with the weekday: "Na terceira segunda-feira de abril" vs "No terceiro domingo de abril"

marcelklehr:
I think the ChattyLLMInputForm is getting slightly out of hand :D Can we refactor it a bit to introduce a new assignments view and new components that are shared between it and the chat view?

marcelklehr:
> I think the ChattyLLMInputForm is getting slightly out of hand :D Can we refactor it a bit to introduce a new assignments view and new components that are shared between it and the chat view?

You don't need to hurry on this one, Julien is also working on the ChattyLLMInputForm atm and would like to merge his changes first.

src/components/ChattyLLM/recurrenceRule.js
marcelklehr: Since calendar also does things this way, we decided to keep it like this. Let's perhaps put a comment here for posterity

marcelklehr:
@copilot resolve the merge conflicts in this pull request

marcelklehr (approved):

marcelklehr:
> Can we refactor it a bit to introduce a new assignments view and new components that are shared between it and the chat view?

Nevermind, let's go with this

PR #559, opened 2026-05-26
Assistant dialog papercuts

### LINKED ISSUE #534: Assistant papercuts NC34
A few papercuts and possible improvements I found while reviewing recent changes, feel free to pick up what you can and/or create sub issues from any of the points.

Image

- [ ] Increase minimum width for displaying the history to 900px.
- [ ] Increase default width of the dialog. Currently it's set at 50% which is unnecessarily narrow on laptop screens. Even if we have a new draggable window we should try to maintain the sizing logic we had for the old dialog (at least for the initial size)
Image

marcoambrosini:
Fix: #534.

- Collapse task history sidebar below 900px dialog width
- Widen the assistant dialog to `min(1220px, 90vw)`
- Fullscreen on small mobile 
- Fullscreen when maximzed (before it wasn't full height)
- Style and position the maximize/minimize button similarly to the close button

Before 

https://github.com/user-attachments/assets/bec29794-e554-478e-9c15-8620b268e6cb

After

https://github.com/user-attachments/assets/0056b376-59fe-40d2-b464-dea6d246779c

### DIFF
--- a/src/components/AssistantTextProcessingModal.vue
+++ b/src/components/AssistantTextProcessingModal.vue
@@ -221,7 +221,7 @@ export default {
 	height: calc(100vh - 32px);
 	max-height: calc(100vh - 32px);
 	height: 80%;
-	width: 70%;
+	width: min(1220px, 90vw);
 	resize: both;
 	overflow: hidden;
 	filter: drop-shadow(0 0 15px rgba(77, 77, 77, 0.5));
--- a/src/components/AssistantTextProcessingModal.vue
+++ b/src/components/AssistantTextProcessingModal.vue
@@ -8,10 +8,10 @@
 		:closable="false"
 		:dismissable-mask="false"
 		:close-on-escape="false"
-		:draggable="true"
+		:draggable="!isSmallMobile"
 		append-to="self"
 		:base-z-index="isInsideViewer ? 9998 : 5000"
-		class="assistant-modal">
+		:class="['assistant-modal', { 'assistant-modal--fullscreen': isSmallMobile }]">
 		<div ref="modal_content"
 			class="assistant-modal--wrapper">
 			<div class="assistant-modal--content">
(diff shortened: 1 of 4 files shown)


kra-mo (approved):

julien-nc (commented):
The button to toggle the left navigation sidebar disappears when the screen is narrow so it's impossible to navigate the task history (or the chat conversation list). Is it intentional?

Otherwise lgtm 👍

marcoambrosini:
Thanks for the review @julien-nc 
 
> it's impossible to navigate the task history

This was actually a bug caused by my css approach. I reworked this a bit and now using a bit of a hack to toggle the navigation open and close. We cannot use the event bus because that would also affect the navigation of the app in the background. So using el.click() 🙈

src/components/AssistantTextProcessingModal.vue
```diff
@@ -72,6 +73,9 @@ export default {
 		NcButton,
 		CloseIcon,
 	},
+	setup() {
+		return { isSmallMobile: useIsSmallMobile() }
+	},
```
julien-nc: This block should go after the `emits` block.

julien-nc (changes requested):
Works well!
This introduces a small blocking issue though:
The `.p-dialog-header` height is now too smal, it does not cover the space taken by its child elements. Not sure which change introduced that.
The result is that the area we can grab to move the dialog is way thinner.

src/components/AssistantTextProcessingModal.vue
marcoambrosini: Nice catch!

src/mixins/navAutoCollapse.js
```diff
@@ -0,0 +1,75 @@
...
+	},
+	methods: {
+		setupNavObserver() {
+			this.teardownNavObserver()
+			const container = this.$refs.container
+			if (!container || typeof ResizeObserver === 'undefined') {
+				return
+			}
+			// null = not evaluated yet, so the first observation always applies
+			this.navCollapsed = null
```
julien-nc, line 43: This data prop should be declared and initialized in data()

julien-nc (changes requested):
Lgtm. One last small change and let's get this in.

jancborchardt (approved):
Looks good to me. :) @julien-nc ?

PR #561, opened 2026-05-28
fix(assignments): Add timezone to assignments

marcelklehr:
plus some tests


julien-nc (approved):
Lgtm

marcelklehr:
- [x] timezone default to user timezone

PR #568, opened 2026-06-02
Use the current date by default for new messages and new sessions

### LINKED ISSUE #500: Make timestamp parameter optional in chat API and default to server-side "time()"
Follow-up from #463

Right now `newSession` and `newMessage` require clients to pass a timestamp. This is error-prone (see #452 where a client sent milliseconds instead of seconds) and redundant - the server already uses `time()` for assistant responses in `ChattyLLMTaskListener`

The timestamp parameter should be made optional and default to `time()` server-side.

julien-nc:
closes #500

Self-explanatory :grin:

### DIFF
--- a/lib/Controller/ChattyLLMController.php
+++ b/lib/Controller/ChattyLLMController.php
@@ -199,7 +199,7 @@ class ChattyLLMController extends OCSController {
 	 *
 	 * Create a new chat session, add a system message with user instructions
 	 *
-	 * @param int $timestamp The session creation date
+	 * @param ?int $timestamp The session creation date
 	 * @param ?string $title The session title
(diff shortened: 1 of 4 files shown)


lib/Controller/ChattyLLMController.php
```diff
@@ -199,7 +199,7 @@ private function improveAgencyActionNames(array $actions): array {
 	 *
 	 * Create a new chat session, add a system message with user instructions
 	 *
-	 * @param int $timestamp The session creation date
+	 * @param ?int $timestamp The session creation date
```
marcelklehr, line 202: Didn't know this was valid docblock syntax

marcelklehr (approved):

PR #574, opened 2026-06-07
Feature/chat message search

annaneliaa:
Added for backend:
- ``MessageMapper::searchMessages()`` a method to query messages across all sessions from a user sessions. It uses case-insensitive LIKE '%query%' search. The messages table is joined against the sessions table to ensure user scoping with the user id. A limitation here is that search does not use a database index on the content column, so a full table scan is performed. On large instances with a lot of message this can become slow.
- ``ChatService::searchMessages()`` a method to orchestrate the search. It returns matching messages with their session IDs
- ``ChattyLLMController::searchMessages()``  new GET /chat/search?query= OCS endpoint
- Route registered in ``appinfo/routes.php``
- Unit tests for the service layer covering empty query, null user, deduplication of session IDs, and multiple session results

Added for frontend:
- ``searchMessages()``an api helper to call GET /chat/search?query= endpoint
- Search input field (imported from ``@nextcloud/vue/components/NcAppNavigationSearch``) in the chat sidebar
- Debounced querying at a 350 ms rate and a clear button
- Session list filters to show only conversations containing matching messages when a search is active
- An info message in the sidebar with "No conversations match your search" for when the query returns no results
- ``searchQuery`` is passed as a prop through ``ChattyLLMInputForm.vue`` to ``ConversationBox.vue`` to ``Message.vue``
- Text matching the search query is highlighted in yellow via ``highlightedContent`` that is computed (+styled) in ``Message.vue``

### DIFF
--- a/lib/Db/ChattyLLM/MessageMapper.php
+++ b/lib/Db/ChattyLLM/MessageMapper.php
@@ -191,7 +191,7 @@ class MessageMapper extends QBMapper {
 	 */
 	public function searchMessages(string $userId, string $query, int $limit = 100): array {
 		$qb = $this->db->getQueryBuilder();
-		$qb->select(Message::$columns)
+		$qb->select(array_map(fn ($col) => 'm.' . $col, Message::$columns))
 			->from($this->getTableName(), 'm')
 			->join('m', 'assistant_chat_sns', 's', $qb->expr()->eq('m.session_id', 's.id'))
 			->where($qb->expr()->eq('s.user_id', $qb->createPositionalParameter($userId, IQueryBuilder::PARAM_STR)))
--- a/lib/Db/ChattyLLM/MessageMapper.php
+++ b/lib/Db/ChattyLLM/MessageMapper.php
@@ -190,14 +190,29 @@ class MessageMapper extends QBMapper {
 	 * @throws \OCP\DB\Exception
 	 */
 	public function searchMessages(string $userId, string $query, int $limit = 100): array {
+		// Split the query into individual words on whitespace
+		$words = array_filter(preg_split('/\s+/', trim($query)));
+
+		if (empty($words)) { // A guard for the edge case where the query is empty or consists of all whitespaces
+			return [];
+		}
+
 		$qb = $this->db->getQueryBuilder();
-		$qb->select(array_map(fn ($col) => 'm.' . $col, Message::$columns))
+		$qb->select(array_map(fn ($col) => 'm.' . $col . ' AS ' . $col, Message::$columns))
 			->from($this->getTableName(), 'm')
-			->join('m', 'assistant_chat_sns', 's', $qb->expr()->eq('m.session_id', 's.id'))
-			->where($qb->expr()->eq('s.user_id', $qb->createPositionalParameter($userId, IQueryBuilder::PARAM_STR)))
-			->andWhere($qb->expr()->iLike('m.content', $qb->createPositionalParameter('%' . $this->db->escapeLikeParameter($query) . '%', IQueryBuilder::PARAM_STR)))
-			->orderBy('m.timestamp', 'DESC')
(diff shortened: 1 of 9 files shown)


lib/Db/ChattyLLM/MessageMapper.php
```diff
@@ -181,4 +181,24 @@ public function deleteMessageById(int $sessionId, int $messageId): void {
...
+	/**
+	 * @param string $userId
+	 * @param string $query
+	 * @param int $limit
+	 * @return list<Message>
+	 * @throws \OCP\DB\Exception
+	 */
+	public function searchMessages(string $userId, string $query, int $limit = 100): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select(array_map(fn ($col) => 'm.' . $col, Message::$columns))
```
julien-nc: The result of this query is passed to `findEntities` which expects original column names. Some databases might keep the prefix (`m.`) in the query result and make it impossible for `findEntities` to build entities. I think it's safer to alias all the selected columns as the original column name.
```suggestion
		$qb->select(array_map(fn ($col) => 'm.' . $col . ' AS ' . $col, Message::$columns))
```
I have not tested my suggested change but it should be possible to do that. Wdyt?

src/components/ChattyLLM/ChattyLLMInputForm.vue
```diff
@@ -392,6 +413,18 @@ export default {
 				this.loading.titleGeneration = false
 			}
 		},
+
+		searchQuery(newVal) {
+			console.debug('[search] searchQuery changed to:', JSON.stringify(newVal))
+			clearTimeout(this.searchDebounceTimer)
```
julien-nc, line 424: Nice that it debounces. I think there still is a race condition producing unexpected results.
In short: If a request is very long, another following request might finish earlier, then the first request will overwrite the second one.
Here is a sequence of actions:
* 1st request running (and takes a long time)
* The user updates the search query
* The UI waits the debounce delay and makes a 2nd request (the first request is still running)
* The 2nd request ends
* The 1st request ends and overwrites the results of the 2nd

The solution for that is to cancel the potentially running previous request whenever the search query changes. You can find examples of how to cancel a request in the assistant's frontend. There is one in this very file. Look for `messagesAxiosController`.
Wdyt?

src/components/ChattyLLM/Message.vue
```diff
@@ -139,6 +143,10 @@ export default {
 			type: Object,
 			default: null,
 		},
+		searchQuery: {
```
julien-nc, line 143: If i understand correctly, the search query is passed to all messages. We could save a bit of rendering effort by just passing it to the messages that are part of the search result (those that match).

src/components/ChattyLLM/Message.vue
```diff
@@ -171,6 +179,17 @@ export default {
 		audioAttachments() {
 			return this.message.attachments?.filter(a => a.type === SHAPE_TYPE_NAMES.Audio) ?? []
 		},
+		highlightedContent() {
```
julien-nc: Nicely done. There is a problem with this approach: The content is not rendered as markdown anymore.

Maybe a simpler approach would be to just wrap the matching content substrings with the bold markdown markers. This would mean replacing all occurrences of the match by `**MATCHED_SUB_STRING**` and the markdown rendering (already done by NcRichText) would do the rendering/styling job naturally.

Wdyt?

julien-nc (changes requested):
Very nice and thorough PR! Thank you.

Here is a review for later, when we will want to push this further and merge. You don't have to follow up now. This is out of scope for the interview process.

janepie:
Nice PR! One small nice-to-have stretch goal: the string has to match exactly right now, so "news brussels" won't find a mesagge that includes "news in Brussels", but we can adapt that later

PR #587, opened 2026-06-22
fix(AssignmentsService): Do not run assignments whose sessions have p…

marcelklehr:
…ending actions

## 🤖 AI (if applicable)


julien-nc (approved):
Nice catch, we didn't think about that.

PR #590, opened 2026-06-26
Make the text input larger by default

lukasdotcom:
This is to invite users to put in more text.
image

## 🤖 AI (if applicable)


kyteinsky (approved):
nice!

kyteinsky:
although not sure if just Enter to submit makes sense now since mutiple lines are expected.
maybe Ctrl+Enter or some combination would be better and can be indicated in the tooltip of the submit button or even more explicitly in the submit button's text.

PR #592, opened 2026-06-28
fix: avoid mime_content_type warning on wrapped streams

### LINKED ISSUE #591: mime_content_type() logs a stream_cast warning on every Assistant preview/output-file request

3.4.3

33.0.6.2

Not browser-specific — this is a server-side PHP warning, reproducible via any client including the OCS API directly (e.g. via curl).

### Describe the Bug

mime_content_type() logs a PHP warning on every Assistant preview/output-file
request:

mime_content_type(): Icewind\Streams\CallbackWrapper::stream_cast is not implemented!

This comes from lib/Service/AssistantService.php, in getTargetFileName()
and getOutputFilePreviewFile(), both of which call:

mime_content_type($file->fopen('rb'))

File::fopen('rb') can return a stream wrapped in
Icewind\Streams\CallbackWrapper, which does not implement stream_cast().
mime_content_type() needs that to inspect the underlying resource, so PHP
falls back to another detection method and logs the warning as a side
effect.

This is log noise, not a functional bug — I confirmed against a live
33.0.6 instance that translation and image-generation tasks both complete
successfully end-to-end with the mime type still correctly detected; the
only symptom is the warning appearing in the log.

The same warning was previously hit in core's TaskProcessingApiController
and addressed in nextcloud/server#54673.

I have a fix ready (drains the stream into a temp file before calling
mime_content_type(), instead of passing the stream resource directly) and
will open a PR referencing this issue shortly.

### Expected Behavior

No warning is logged. The task completes and the mime type is still
correctly detected (this already works today — only the log noise should
be eliminated).

### To Reproduce

1. Run any Assistant task that produces an output file (e.g. text-to-image
   generation), or request a preview of an output file via the OCS API:
   GET .../assistant/api/v1/task/{id}/output-file/{id}/preview
2. Check data/nextcloud.log (or your configured log target) for entries
   logged during that request.
3. Observe the mime_content_type() / stream_cast warning.

olddude2:
mime_content_type() requires stream_cast() support to operate on a stream resource directly. Streams returned by File::fopen() can be wrapped in Icewind\Streams\CallbackWrapper, which does not implement stream_cast(), causing a PHP warning to be logged on every preview/output-file request even though the mime type is still detected correctly.

This drains the stream into a temporary file first, then calls mime_content_type() against the file path instead. No behavior change; only removes log noise.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Service/AssistantService.php
+++ b/lib/Service/AssistantService.php
@@ -743,7 +743,8 @@ class AssistantService {
 	 * @throws NotPermittedException
 	 */
 	private function getTargetFileName(File $file): string {
-		$mimeType = $this->detectMimeType($file->fopen('rb'));
+		$head = fread($file->fopen('rb'), 4096);
+		$mimeType = (new \finfo(FILEINFO_MIME_TYPE))->buffer($head);
 		$fileName = $file->getName();
 
 		$mimes = new \Mimey\MimeTypes;
@@ -753,29 +754,6 @@ class AssistantService {
 			return $fileName . '.' . $extension;
 		}
 		return $fileName;
-	}
-
-	/**
-	 * Wraps mime_content_type() to avoid a PHP warning when given a stream
-	 * that does not implement stream_cast() (e.g. Icewind\Streams\CallbackWrapper,
-	 * returned by File::fopen() for some storage backends). The stream is
-	 * drained into a temp file first so mime_content_type() can operate on a
-	 * real path instead of the unsupported stream resource.
-	 *
-	 * @param resource $stream
-	 * @return string|false
-	 */
-	private function detectMimeType($stream) {
-		$tmpFile = tempnam(sys_get_temp_dir(), 'nc_assistant_mime_');
-		$tmpHandle = fopen($tmpFile, 'wb');
-		stream_copy_to_stream($stream, $tmpHandle);
-		fclose($tmpHandle);
-		if (is_resource($stream)) {
-			fclose($stream);
-		}
-		$mimeType = mime_content_type($tmpFile);
-		unlink($tmpFile);
-		return $mimeType;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


olddude2:
The PHPUnit / php8.2-sqlite-master failure appears to be a pre-existing 
issue — the job fails during Nextcloud installation before any tests run, 
with "This version of Nextcloud requires at least PHP 8.3" (PHP 8.2.31 is 
being used against master). The php8.2-sqlite-stable33 job passes cleanly.

Happy to rebase or make any adjustments if needed — just let me know.

lib/Service/AssistantService.php
```diff
@@ -807,7 +830,7 @@ private function extractFileIdsFromTask(Task $task): array {
 	 */
 	public function getOutputFilePreviewFile(string $userId, int $taskId, int $fileId, ?int $x = 100, ?int $y = 100): ?array {
 		$taskOutputFile = $this->getTaskOutputFile($userId, $taskId, $fileId);
-		$realMime = mime_content_type($taskOutputFile->fopen('rb'));
+		$realMime = $this->detectMimeType($taskOutputFile->fopen('rb'));
```
janepie:
```suggestion
        $head = fread($stream, 4096);
        if (is_resource($stream)) {
            fclose($stream);
        }
       $realMime = new \finfo(FILEINFO_MIME_TYPE)->buffer($head);

```

Thank you for your contribution! Unfortunately we cannot merge this as copying the whole file to disk has a too high performance impact especially for big files. An easier fix would be to not use mime_content_type as shown above, would you be up for testing and validating if that works without any issues?

janepie:
fixes #591

olddude2:
Hi — just checking in on this whenever you get a chance. I updated the fix per the earlier feedback to use finfo::buffer() on a small read instead of the temp-file approach. Let me know if anything else needs adjusting. Thanks for your time on this. Cheers!

janepie (approved):
Sorry for coming back on this so late. Looks good, will be included in the NC35 release!

PR #594, opened 2026-06-30
feat: add context agent skills endpoints and admin settings

kyteinsky:
related: https://github.com/nextcloud/context_agent/pull/206

The folder selection dialog lists the current admin's files. Once selected, the path and the user id are stored. This screenshot is when the admin user was logged in and the folder was set from alice's filesystem.

image

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI


lib/Service/AgentSkillsService.php
```diff
@@ -0,0 +1,396 @@
...
+	 *                                 existing one was replaced
+	 * @throws \InvalidArgumentException if the skill name is empty or contains a slash
+	 * @throws NotFoundException if the skills folder cannot be resolved
+	 * @throws NotPermittedException if the skill folder or SKILL.md file cannot be written, or if a
+	 *                               non-folder node already exists at the target skill path
+	 * @throws \OC\User\NoUserException if the user does not exist
+	 * @throws \OCP\Files\GenericFileException if writing the SKILL.md file fails
+	 * @throws \OCP\Lock\LockedException if the SKILL.md file is locked
+	 */
+	public function storeSkill(string $userId, string $skillName, string $description, string $content): string {
```
marcelklehr, line 198: This should enforce the same rules as the tools: https://github.com/nextcloud/context_agent/pull/206/changes#diff-433032706a1167f61410a08c64c8739e25a62cd2af63ec271606757395730857R21-R23

kyteinsky: I thought about giving some leeway to external tools using the skills feature but sure, let's enforce the rules since it's widely accepted anyway.

marcelklehr (approved):

lib/Service/AgentSkillsService.php
```diff
@@ -0,0 +1,396 @@
...
+		if ($skillName === '' || str_contains($skillName, '/')) {
```
julien-nc: We could be more defensive here and reject all the chars that are forbidden by NC.
Hmmm anyway `newFile` will throw if a forbidden char is used. Wdyt?

lib/Service/AgentSkillsService.php
```diff
@@ -0,0 +1,396 @@
...
+		$content = $file->getContent();
+		$delimiter = self::FRONTMATTER_DELIMITER;
+
+		// must start with the opening delimiter followed by a newline
+		if (!str_starts_with($content, $delimiter . "\n") && !str_starts_with($content, $delimiter . "\r\n")) {
+			throw new RuntimeException('Skill file missing frontmatter opening delimiter: ' . $file->getPath());
+		}
+
+		$offset = strpos($content, "\n") + 1;
+		$closingPos = strpos($content, "\n" . $delimiter, $offset);
```
julien-nc: This could match something inside the frontmatter block.
Is it allowed to have something like:
```
---
hey ho
---plop
more content
---
```

If so, we should make sure we detect a closing delimiter that matches `\n---$`.

lib/Service/AgentSkillsService.php
```diff
@@ -0,0 +1,396 @@
...
+			} catch (NotFoundException $e) {
+				$this->logger->debug('Skipping skill folder without ' . self::SKILL_FILE_NAME . ': ' . $node->getName());
+				continue;
+			}
+			if (!$skillFile instanceof File) {
+				continue;
+			}
+			try {
+				$skills[$node->getName()] = $this->getSkillMetadata(
+					$skillCacheKeyPrefix . md5($node->getName()),
```
julien-nc: We could use something with less collision risk than md5 like `hash('sha256', $node->getName())`. Wdyt?

lib/Service/AgentSkillsService.php
```diff
@@ -0,0 +1,396 @@
...
+	 * @throws NotPermittedException if the skills folder cannot be created
+	 * @throws \OC\User\NoUserException if the user does not exist
+	 */
+	private function getSkillsFolder(string $userId): Folder {
+		$assistantFolder = $this->assistantService->getAssistantDataFolder($userId);
+		$skillsFolderPath = self::SKILLS_FOLDER_PATH;
+
+		if ($assistantFolder->nodeExists($skillsFolderPath)) {
+			$node = $assistantFolder->get($skillsFolderPath);
+			if ($node instanceof Folder) {
```
julien-nc, line 390: I think there a case that is not handled: When the `$skillsFolderPath` node exists but is a file, we will still try to create a directory with the same name.

julien-nc (changes requested):
* Do we need a DELETE endpoint?

lib/Service/AgentSkillsService.php
kyteinsky: good idea and early validation is still better, will do!

lib/Service/AgentSkillsService.php
kyteinsky: nice catch

julien-nc (approved):

PR #595, opened 2026-06-30
feat: Capture and display reasoning content

marcelklehr:
## 🤖 AI (if applicable)

### DIFF
--- a/src/components/ChattyLLM/Message.vue
+++ b/src/components/ChattyLLM/Message.vue
@@ -35,7 +35,8 @@
 					<NcPopover v-if="message.reasoning">
 						<template #trigger>
 							<NcButton
-								:aria-label="t('assistant', 'Reasoning content')">
+								:aria-label="t('assistant', 'Reasoning content')"
+								:title="t('assistant', 'Reasoning content')">
 								<template #icon>
 									<ReasoningContentIcon :size="20" />
 								</template>
@@ -53,7 +54,8 @@
 					<NcPopover v-if="parsedSources.length && !streaming">
 						<template #trigger>
 							<NcButton
-								:aria-label="t('assistant', 'Information sources & actions')">
+								:aria-label="t('assistant', 'Information sources & actions')"
+								:title="t('assistant', 'Information sources & actions')">
 								<template #icon>
 									<ToolInformationIcon :size="20" />
 								</template>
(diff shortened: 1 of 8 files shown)


src/components/ChattyLLM/Message.vue
```diff
@@ -247,6 +269,9 @@ export default {
 			}
 			return this.informationSourceNames[source] ? this.informationSourceNames[source] : source
 		},
+		toggleReasoning() {
+			this.showReasoning = !this.showReasoning
+		},
```
julien-nc: This is not used.

src/components/ChattyLLM/Message.vue
```diff
@@ -31,19 +31,37 @@
 				<div class="message__header__role__name">
 					{{ message.role === 'human' ? displayName : t('assistant', 'Nextcloud Assistant') }}
 				</div>
-				<div style="display: flex">
+				<div style="display: flex; gap: 5px">
+					<NcPopover v-if="message.reasoning && !streaming">
```
julien-nc:
```suggestion
					<NcPopover v-if="message.reasoning">
```
If we want to stream the reasoning content.
Line 6 needs `|| message.reasoning` to display the message if there is some reasoning content.

julien-nc (changes requested):
Right now the message (or the placeholder animation) is not displayed when there is some reasoning content and we are streaming (and there is no content yet).

Not sure we want to stream the reasoning. I think it's nice but not mandatory. If we want it, there are some minimal changes to make it happen.

src/components/icons/aliases.js
```diff
@@ -0,0 +1,11 @@
+<!--
+  - SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
```
julien-nc:
```suggestion
/**
 * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
 * SPDX-License-Identifier: AGPL-3.0-or-later
 */
```
Can't build otherwise.

julien-nc (approved):
Works great. Mandatory change, the frontend stuff can't be built right now.

src/components/icons/aliases.js
marcelklehr: Strange that I didn't get that error.

julien-nc:
You can just use Php 8.3 in the Phpunit action to fix CI.

PR #596, opened 2026-06-30
Add improve button to text outputs to allow for iteration

lukasdotcom:
Example in generate text:
image

It is on _all_ text output fields right now. If you can think of some it shouldn't be on feel free to mention that.

## 🤖 AI (if applicable)

### DIFF
--- a/src/components/AssistantTextProcessingForm.vue
+++ b/src/components/AssistantTextProcessingForm.vue
@@ -123,7 +123,9 @@
 							v-model:show-advanced="showAdvanced"
 							v-model:outputs="myOutputs"
 							:inputs="myInputs"
-							:selected-task-type="selectedTaskType" />
+							:selected-task-type="selectedTaskType"
+							:can-improve-output="canImproveOutput"
+							@improve="onImprove" />
 					</div>
 					<div class="footer">
 						<div class="footer--action-buttons">
@@ -254,24 +256,6 @@ export default {
 		return {
 			providedCurrentTaskId: () => this.selectedTaskId,
 			streaming: () => this.streaming,
-			improveOutput: (content) => {
-				if (!this.showImproveWithNewInstructionsButton) {
-					return
-				}
-				this.$emit('new-task')
-				this.mySelectedTaskTypeId = TEXT2TEXT_IMPROVE_TASK_TYPE_ID
-				this.$nextTick(() => {
-					this.$refs?.assistantFormInputs?.setDefaultValues(true)
-					this.$nextTick(() => {
-						this.myInputs = {
-							...this.myInputs,
-							input: content,
-						}
-						saveLastSelectedTaskType(this.mySelectedTaskTypeId)
-					})
-				})
-			},
-			canImproveOutput: () => this.showImproveWithNewInstructionsButton,
 		}
 	},
 	props: {
@@ -479,7 +463,7 @@ export default {
 		actionButtonsToShow() {
 			return this.hasOutput ? this.actionButtons : []
 		},
-		contextWriteTaskType() {
+		improveTaskType() {
 			const taskType = this.taskTypes.find(tt => tt.id === TEXT2TEXT_IMPROVE_TASK_TYPE_ID)
 			if (taskType === undefined) {
 				return null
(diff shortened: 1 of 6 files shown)


marcelklehr (approved):

marcelklehr:
The button could read "Improve this text", so it's more clear what it does, perhaps? Otherwise, improve could also mean something like "help improve the assistant", etc.

src/components/AssistantTextProcessingForm.vue
```diff
@@ -460,6 +479,20 @@ export default {
 		actionButtonsToShow() {
 			return this.hasOutput ? this.actionButtons : []
 		},
+		contextWriteTaskType() {
```
julien-nc:
```suggestion
		improveTaskType() {
```
Not related with ContextWrite, right?

src/components/AssistantTextProcessingForm.vue
```diff
@@ -253,6 +254,24 @@ export default {
 		return {
 			providedCurrentTaskId: () => this.selectedTaskId,
 			streaming: () => this.streaming,
+			improveOutput: (content) => {
```
julien-nc: IMO this should be done with an event (still listen and handle the event here), not by providing a callback to the child components. Wdyt?

julien-nc (changes requested):

src/components/AssistantTextProcessingForm.vue
lukasdotcom: Copypasta I originally wrote it to use context write for testing before I made the improve text task type.

src/components/AssistantTextProcessingForm.vue
lukasdotcom: It seemed easier to not use events because I would need to pass them through all the components, but I've changed it now to use events.

lukasdotcom:
> The button could read "Improve this text", so it's more clear what it does, perhaps? Otherwise, improve could also mean something like "help improve the assistant", etc.

Makes sense

src/components/AssistantTextProcessingForm.vue
```diff
@@ -484,6 +501,23 @@ export default {
...
+			this.$emit('new-task')
+			this.mySelectedTaskTypeId = TEXT2TEXT_IMPROVE_TASK_TYPE_ID
+			this.$nextTick(() => {
+				this.$refs?.assistantFormInputs?.setDefaultValues(true)
+				this.$nextTick(() => {
+					this.myInputs = {
+						...this.myInputs,
+						input: content,
+					}
+					saveLastSelectedTaskType(this.mySelectedTaskTypeId)
```
kyteinsky: nit: improve maybe should not be saved since it would be a second step most of the time from other task types, or is it generic enough to be pre-selected?

src/components/fields/TextInput.vue
```diff
@@ -22,23 +22,33 @@
...
-			</span>
-		</NcButton>
+		<div v-if="isOutput && hasValue"
+			class="output-buttons">
+			<NcButton v-if="!streaming() && canImproveOutput"
+				class="improve-button"
+				variant="secondary"
+				:title="t('assistant', 'Improve with new instructions')"
+				@click="$emit('improve', formattedValue)">
+				{{ t('assistant', 'Improve this text') }}
```
kyteinsky, line 35: would be nice to have an icon like the Copy button, maybe a wrench or pencil, or something similar.

src/components/AssistantTextProcessingForm.vue
lukasdotcom: It technically is generic enough to be used by itself, but I do agree not preselecting it is a good idea.

src/components/fields/TextInput.vue
lukasdotcom: Added wrench icon and updated screenshot

src/components/AssistantTextProcessingForm.vue
julien-nc: You could use the Nextcloud event bus so the event does not have to be forwarded by every component.

src/components/fields/TextInput.vue
kyteinsky: sorry being being nitpicky but the icon looks a bit large compared against the text
this is at size 20 for both improve and copy:
image

size 18 (which looks a bit small):
image

kyteinsky (approved):
🚀

julien-nc (approved):

PR #601, opened 2026-07-08
Audio translation

julien-nc:
* Add an audio translation provider that runs 3 sub tasks (stt, translate, tts)
* Proxy the optional input shape of the underlying translate and tts providers (if there are some known ones)
* Bump min NC to 35
* Set Php 8.3 in config.platform
* Update composer dependencies

## 🤖 AI (if applicable)

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -39,7 +39,6 @@ use OCA\Assistant\TaskProcessing\TextToStickerTaskType;
 use OCA\Files\Event\LoadAdditionalScriptsEvent;
 use OCP\AppFramework\App;
 use OCP\AppFramework\Bootstrap\IBootContext;
-
 use OCP\AppFramework\Bootstrap\IBootstrap;
 use OCP\AppFramework\Bootstrap\IRegistrationContext;
 use OCP\AppFramework\Http\Events\BeforeTemplateRenderedEvent;
@@ -120,10 +119,7 @@ class Application extends App implements IBootstrap {
 		// $context->registerTaskProcessingTaskType(ImageToTextTranslateTaskType::class);
 		// $context->registerTaskProcessingProvider(ImageToTextTranslateProvider::class);
 
-		// AudioToAudioTranslate appeared in 35
-		if (class_exists('OCP\\TaskProcessing\\TaskTypes\\AudioToAudioTranslate')) {
-			$context->registerTaskProcessingProvider(AudioToAudioTranslateProvider::class);
-		}
+		$context->registerTaskProcessingProvider(AudioToAudioTranslateProvider::class);
 	}
 
 	public function boot(IBootContext $context): void {
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -30,6 +30,7 @@ use OCA\Assistant\Reference\TaskOutputFileReferenceProvider;
 use OCA\Assistant\Reference\Text2ImageReferenceProvider;
 use OCA\Assistant\Reference\Text2StickerProvider;
 use OCA\Assistant\TaskProcessing\AudioToAudioChatProvider;
+use OCA\Assistant\TaskProcessing\AudioToAudioTranslateProvider;
 use OCA\Assistant\TaskProcessing\ContextAgentAudioInteractionProvider;
 use OCA\Assistant\TaskProcessing\ImageToTextTranslateProvider;
 use OCA\Assistant\TaskProcessing\ImageToTextTranslateTaskType;
@@ -118,6 +119,11 @@ class Application extends App implements IBootstrap {
 		// not ready yet
 		// $context->registerTaskProcessingTaskType(ImageToTextTranslateTaskType::class);
 		// $context->registerTaskProcessingProvider(ImageToTextTranslateProvider::class);
+
+		// AudioToAudioTranslate appeared in 35
+		if (class_exists('OCP\\TaskProcessing\\TaskTypes\\AudioToAudioTranslate')) {
+			$context->registerTaskProcessingProvider(AudioToAudioTranslateProvider::class);
+		}
 	}
 
 	public function boot(IBootContext $context): void {
--- a/lib/TaskProcessing/AudioToAudioTranslateProvider.php
+++ b/lib/TaskProcessing/AudioToAudioTranslateProvider.php
@@ -73,9 +73,17 @@ class AudioToAudioTranslateProvider implements IProvider, ISynchronousOptionsAwa
 	}
 
 	public function getOptionalInputShape(): array {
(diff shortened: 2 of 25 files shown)


lib/AppInfo/Application.php
```diff
@@ -118,6 +118,11 @@ public function register(IRegistrationContext $context): void {
 		// not ready yet
 		// $context->registerTaskProcessingTaskType(ImageToTextTranslateTaskType::class);
 		// $context->registerTaskProcessingProvider(ImageToTextTranslateProvider::class);
+
+		// AudioToAudioTranslate appeared in 35
+		if (class_exists('OCP\\TaskProcessing\\TaskTypes\\AudioToAudioTranslate')) {
```
lukasdotcom: Not necessary as you bumped the minimum version

lukasdotcom (commented):
Didn't test this, but looks fine.

lib/AppInfo/Application.php
julien-nc: Very true. Removed.

lib/AppInfo/Application.php
```diff
@@ -118,6 +118,8 @@ public function register(IRegistrationContext $context): void {
 		// not ready yet
 		// $context->registerTaskProcessingTaskType(ImageToTextTranslateTaskType::class);
 		// $context->registerTaskProcessingProvider(ImageToTextTranslateProvider::class);
+
+		$context->registerTaskProcessingProvider(AudioToAudioTranslateProvider::class);
```
marcelklehr, line 122: Should we only register this if all subordinate task types have a provider available? Otherwise getPreferredTaskType throws

lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,339 @@
...
+			if ($ttsModel !== null) {
+				$ttsInput['model'] = $ttsModel;
+			}
+			if ($ttsVoice !== null) {
+				$ttsInput['voice'] = $ttsVoice;
+			}
+			if ($ttsSpeed !== null) {
+				$ttsInput['speed'] = $ttsSpeed;
+			}
+			// this provider is not declared if TextToSpeech does not exist so we know it's fine
```
marcelklehr: Not true, I think (see Application.php)

lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,339 @@
...
+			/** @psalm-suppress UndefinedClass */
+			$task = new Task(
+				TextToSpeech::ID,
+				$ttsInput,
+				Application::APP_ID . ':internal',
+				$userId,
+			);
+			$task->setIncludeWatermark(true);
```
marcelklehr: Why is this true here? Should we use getIncludeWatermark() from the otpions instead of hardcoding?

lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,339 @@
...
+	public function getOptionalInputShapeEnumValues(): array {
+		$translateProvider = $this->taskProcessingService->getPreferredProvider(TextToTextTranslate::ID);
+		$translateOptionalInputShape = $translateProvider->getOptionalInputShape();
+		$ttsProvider = $this->taskProcessingService->getPreferredProvider(TextToSpeech::ID);
+		$ttsOptionalInputShape = $ttsProvider->getOptionalInputShape();
+
+		$optionalInputShapeEnumValues = [];
+
+		if (isset($translateOptionalInputShape['model'])) {
+			$optionalInputShapeEnumValues['translation_model'] = $translateProvider->getOptionalInputShapeEnumValues()['model'];
```
marcelklehr: `$translateProvider->getOptionalInputShapeEnumValues()['model']` might not be set, even when `$translateOptionalInputShape['model']` is set

julien-nc: Fixed

lib/TaskProcessing/AudioToAudioTranslateProvider.php
julien-nc: Fixed

lib/TaskProcessing/AudioToAudioTranslateProvider.php
julien-nc: Remaining of a copy pasta. Removed the psalm-suppress and the comment.

lib/AppInfo/Application.php
julien-nc: I tried 
``` php
$availableTaskTypeIds = $this->taskProcessingManager->getAvailableTaskTypeIds();
if (in_array(AudioToText::ID, $availableTaskTypeIds, true)) {
	$context->registerTaskProcessingProvider(AudioToAudioTranslateProvider::class);
}
```
but at this point not all the task types are registered, on my setup I only get `core:audio2audio:chat` and `assistant:text2sticker`. Not sure if we can really check what's available here.

lib/TaskProcessing/AudioToAudioTranslateProvider.php
```diff
@@ -0,0 +1,343 @@
...
+
+	public function getInputShapeDefaults(): array {
+		$translateProvider = $this->taskProcessingService->getPreferredProvider(TextToTextTranslate::ID);
+		return [
+			'origin_language' => $translateProvider->getInputShapeDefaults()['origin_language'],
+		];
+	}
+
+	public function getOptionalInputShape(): array {
+		$translateProvider = $this->taskProcessingService->getPreferredProvider(TextToTextTranslate::ID);
```
marcelklehr: > at this point not all the task types are registered, on my setup I only get core:audio2audio:chat and assistant:text2sticker. Not sure if we can really check what's available here.

Then we should catch errors in getPreferredProvider, I think? Or what happens currently when the subordinate task providers are not available?

julien-nc: When no `TextToSpeech` provider is registered, `getPreferredProvider(TextToSpeech::ID)` throws a `OCP\TaskProcessing\Exception\Exception` that is caught in `Manager::getAvailableTaskTypes` and the `AudioToAudioTranslateProvider` provider is skipped. This is not clean but has no bad side effect and behaves like we want it to: Audio translation does not appear in the task type list in the settings of in the assistant dialog.

We could still catch this exception and throw a better one with a message like `Missing required provider for task type TT`. Wdyt?

julien-nc: I did that in the last commit.

marcelklehr (approved):

PR #602, opened 2026-07-09
feat: support multimodal chat

lukasdotcom:
Adds support for multimodal chat.
image

Needs these prs:
- https://github.com/nextcloud/integration_openai/pull/400
- https://github.com/nextcloud/server/pull/61963
You can also now use llm2 instead
- https://github.com/nextcloud/llm2/pull/278

For context agent you can use this pr:
- https://github.com/nextcloud/context_agent/pull/211

## 🤖 AI (if applicable)

- [X] The content of this PR was partly or fully generated using AI

### DIFF
--- a/lib/Service/ChatService.php
+++ b/lib/Service/ChatService.php
@@ -845,10 +845,7 @@ class ChatService {
 			'confirmation' => $confirmation,
 			'conversation_token' => $conversationToken,
 		];
-		/** @psalm-suppress UndefinedClass */
-		if (isset($this->taskProcessingManager->getAvailableTaskTypes()[\OCP\TaskProcessing\TaskTypes\MultimodalContextAgentInteraction::ID]['optionalInputShape']['memories'])) {
-			$taskInput['memories'] = $this->sessionSummaryService->getMemories($userId);
-		}
+		$taskInput['memories'] = $this->sessionSummaryService->getMemories($userId);
 		/** @psalm-suppress UndefinedClass */
 		$task = new Task(
 			\OCP\TaskProcessing\TaskTypes\MultimodalContextAgentInteraction::ID,
--- a/lib/Service/ChatService.php
+++ b/lib/Service/ChatService.php
@@ -447,39 +447,43 @@ class ChatService {
 				$fileId = $audioAttachment['file_id'];
 				$taskId = $this->scheduleAudioChatTask($userId, $fileId, $systemPrompt, $history, $sessionId, $lastUserMessage->getId());
 			} else {
-				// for a text chat task, let's only use text in the history
-				$historyMessages = array_map(static function (Message $message) {
-					return json_encode([
-						'role' => $message->getRole(),
-						'content' => $message->getContent(),
-					]);
-				}, $history);
 				if ($this->isMultimodalChatAvailable()) {
-					// for a multimodal chat task, let's use the attachments in the history as we don't know the structure in history
+					// for a multimodal chat also attachments need to be added to the history
 					$assistantService = $this->assistantService;
-					$attachmentsHistory = array_map(static function (Message $message) use ($userId, $assistantService) {
-						// Make sure that the user has access to the output files
-						if ($message->getRole() === Message::ROLE_ASSISTANT) {
-							$attachments = $message->jsonSerialize()['attachments'];
-							$result = [];
-							foreach ($attachments as $attachment) {
+					$historyMessages = array_map(static function (Message $message) use ($userId, $assistantService) {
+						$attachments = $message->jsonSerialize()['attachments'];
+						// Attachments that were generated need to be saved in the user's files so they are ac
+						$content = array_map(static function (array $attachment) use ($userId, $assistantService, $message) {
+							if ($message->getRole() === Message::ROLE_ASSISTANT) {
 								$info = $assistantService->saveOutputFile($userId, $message->getOcpTaskId(), $attachment['file_id']);
-								$result[] = [
-									'type' => 'File',
+								return [
+									'type' => 'file',
 									'file_id' => $info['fileId'],
 								];
 							}
-							return $result;
-						}
-						return $message->jsonSerialize()['attachments'];
+							return ['type' => 'file', 'file_id' => $attachment['file_id']];
+						}, $attachments);
+						$content[] = [
+							'type' => 'text',
+							'text' => $message->getContent(),
+						];
+						return json_encode([
+							'role' => $message->getRole(),
+							'content' => $content,
+						]);
 					}, $history);
-					// Just add all the attachments in the history as attachments for the latest message as we can't know the structure in history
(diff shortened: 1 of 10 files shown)


julien-nc (commented):
* It works very well 👍 
* The message attachments could be in a horizontal layout, right now they take a lot of space

Image

* When there are multiple attachments, the viewer does not open when clicking one attachment. It works when there is only one attachment.

I'll make code reviews next week.

lukasdotcom:
> * It works very well 👍
> 
>     * The message attachments could be in a horizontal layout, right now they take a lot of space
> 
> 
> Image
> 
>     * When there are multiple attachments, the viewer does not open when clicking one attachment. It works when there is only one attachment.
> 
> 
> I'll make code reviews next week.

The viewer didn't open for input attachments only output attachments. I've switched it to now work for all of them and also made the attachments work horizontally
image

lib/Service/ChatService.php
```diff
@@ -40,6 +40,7 @@ public function __construct(
 		private readonly IManager $taskProcessingManager,
 		private readonly LoggerInterface $logger,
 		private readonly ITimeFactory $timeFactory,
+		private readonly AssistantService $assistantService,
```
julien-nc: Not used.

src/components/ChattyLLM/Message.vue
```diff
@@ -273,6 +290,39 @@ export default {
...
+
+			if (this.isOutput) {
+				const url = generateOcsUrl('/apps/assistant/api/v1/task/{taskId}/file/{fileId}/save', {
+					taskId: file.ocp_task_id ?? this.message.ocp_task_id,
+					fileId: file.file_id,
+				})
+				return axios.post(url).then(response => {
+					const savedPath = response.data.ocs.data.path
+					console.debug('[assistant] view output file', savedPath)
+					OCA.Viewer.open({ path: savedPath })
```
julien-nc:
```suggestion
					OCA?.Viewer?.open({ path: savedPath })
```

julien-nc (approved):

src/components/ChattyLLM/InputArea.vue
```diff
@@ -99,14 +151,30 @@ export default {
...
+			picker: (callback) => getFilePickerBuilder(t('assistant', 'Choose a file'))
+				.setMultiSelect(true)
+				.allowDirectories(false)
+				.addButton({
+					id: 'choose-input-file',
+					label: t('assistant', 'Choose'),
+					variant: 'primary',
+					callback: callback(),
+				})
+				.build(),
```
kyteinsky, line 177: nit: not sure if it's possible to limit the mimetypes here since it cannot be judged for every provider, but we may still limit it to images, videos, audio, and text (including office and pdf) files so files like .zip are not shown as selectable, but as said it may be too much work for little gain since it would fail eventually anyway.

src/components/ChattyLLM/InputArea.vue
```diff
@@ -152,10 +227,54 @@ export default {
...
+		onUploadFileSelected() {
+			this.uploadFiles(this.$refs.fileInput.files)
+				.finally(() => {
+					this.$refs.fileInput.value = ''
+				})
+		},
+		uploadFiles(files) {
+			if (!files || files.length === 0) {
+				return Promise.resolve()
+			}
```
kyteinsky, line 247: the `this.attachedFileIds.length > MAX_ATTACHED_FILES` check may be better here (or in `pickerSubmitted`) so we don't upload the files at all if the limit exceeds, and a warning/error message can be sent to the user stating what failed instead of just disabling the send button.
a file size limit for all files might also make sense, although generous, it will prevent upload of very large files weighing 2 GB for example.

src/components/ChattyLLM/InputArea.vue
```diff
@@ -131,6 +199,13 @@ export default {
 	},
 
 	methods: {
+		onPaste(e) {
+			if (!this.multimodalChatAvailable || e.clipboardData.files.length === 0) {
+				return
+			}
+			e.preventDefault()
+			this.uploadFiles(e.clipboardData.files)
+		},
```
kyteinsky, line 208: noice!

lib/Service/ChatService.php
```diff
@@ -446,14 +453,40 @@ public function scheduleMessageGeneration(?string $userId, int $sessionId, int $
...
+							$newAttachment = ['type' => 'file', 'file_id' => $attachment['file_id']];
+							if (isset($attachment['ocp_task_id'])) {
+								$newAttachment['ocp_task_id'] = $attachment['ocp_task_id'];
+							}
+							return $newAttachment;
+						}, $attachments);
+						$content[] = [
+							'type' => 'text',
+							'text' => $message->getContent(),
+						];
```
kyteinsky, line 470: `'role' => 'user'` maybe since this is the user message

kyteinsky (commented):
multimodal, let's gooo!

a few nits that could be ironed out:

nit: creating a new chat with the attachments selected does not clear them up, not sure if it's intentional (since they are already uploaded so would save that for the new chat) but was something unexpected.
maybe a "Clear All" button for the attachments can act as a middle ground keeping the above behaviour.

nit: maybe the filename could be truncated here, or hidden even and can be seen when hovered

Image

lukasdotcom:
> multimodal, let's gooo!
> 
> a few nits that could be ironed out:
> 
> nit: creating a new chat with the attachments selected does not clear them up, not sure if it's intentional (since they are already uploaded so would save that for the new chat) but was something unexpected. maybe a "Clear All" button for the attachments can act as a middle ground keeping the above behaviour.
> 
> nit: maybe the filename could be truncated here, or hidden even and can be seen when hovered
> Image

Took me a bit longer than I wanted to to look at your feedback, but I removed the filename and added it as a title attribute to the image instead. There is also a clear all attachments button now
image

lib/Service/ChatService.php
lukasdotcom: This is not a message even though it looks like one. It is actually part of content which doesn't have a role: https://developers.openai.com/api/reference/resources/chat/subresources/completions/methods/create#(resource)%20chat.completions%20%3E%20(model)%20chat_completion_content_part_text%20%3E%20(schema)
A message looks something like this where the line of code is adding things in the content that doesn't need a role.
```json
{
  "role": "user",
  "content": [
    {"type":"text","text":"Hello"}
  ]
}
```

src/components/ChattyLLM/InputArea.vue
lukasdotcom: I think it is better to just rely on UserfacingExceptions as I did make them (at least for openai integration llm2 still needs them) and they mention the mimetype that doesn't work.

src/components/ChattyLLM/InputArea.vue
lukasdotcom: Added a cap to the file selection and a cap to file size upload.

PR #604, opened 2026-07-15
feat: Delete chat messages and assignments when user is deleted

janepie:


lib/Listener/UserDeletedListener.php
```diff
@@ -0,0 +1,51 @@
...
+
+	public function handle(Event $event): void {
+		if (!($event instanceof UserDeletedEvent)) {
+			return;
+		}
+
+		$userId = $event->getUid();
+
+		try {
+			$this->assignmentsService->deleteAllForUser($userId);
```
julien-nc, line 42: We don't cleanup the jobs created by `SessionSummaryService`. I think we should check if there are some jobs scheduled for the user and remove them from the bg job queue.

lib/Service/AssignmentsService.php
```diff
@@ -84,13 +86,31 @@ public function createAssignment(?string $userId, string $title, string $prompt,
...
+
 	/**
 	 * @throws InternalException|UnauthorizedException
 	 */
 	public function runDueAssignmentsForUser(?string $userId): void {
 		if ($userId === null) {
 			throw new UnauthorizedException();
 		}
+		if ($this->userManager->get($userId) === null) {
+			$this->deleteAllForUser($userId);
```
julien-nc, line 111: Should we also clean everything up here? The chat sessions+messages and the summary bg jobs?

julien-nc:
@lukasdotcom Wdyt of my review comments?

lib/Service/AssignmentsService.php
janepie: What do you mean, they should already be cleaned up via listener, right?

julien-nc: If some users where deleted before this PR, there can still be data hanging around.

janepie: Fair, but maybe that should be a migration step and not solved here?

julien-nc: Yep, a repair step seems appropriate.

lukasdotcom: A repair step makes sense for this.

lib/Listener/UserDeletedListener.php
lukasdotcom: Pushed a new commit that does both.

julien-nc (approved):
👍

PR #606, opened 2026-07-15
 [stable34] feat: Delete chat messages and assignments when user is deleted 

janepie:
No assignments in stable34 so I had to backport manually


julien-nc (approved):
Because there's a migration in this PR, it's safer to bump the app version number to make sure it will get triggered. Not a big deal because we are sure the next release will have a version bump. It's just in case someone checks out this branch.

lukasdotcom (approved):
Tested this manually on a Nextcloud 34 instance and it deleted messages, sessions and the background job for the summary.

Issue #497, opened 2026-03-25
Syntax highlighting in Chat


derStephan:
### Describe the feature you'd like to request

Markdown has syntax highlighting features like
````
```php
<the code goes here>
```
````

This is working in other apps like notes. However, this does not work in assistent chats sessions. The same applies for tables. 

BTW: to force your AI assistent to always use that, just add `Enclose every code example in triple back–ticks with syntax highlighting.` to the Chat User Instructions for Chat Completions.

### Describe the solution you'd like

Use the same mark down interpreter throughout all components to have the same feature set everywhere

### Describe alternatives you've considered

none, sorry.

janepie:
Thanks for suggesting this! We'll look into it

janepie:
We are not planning to add this on our own but would be happy to help with a PR for this :)

dtump:
@derStephan #613 (just merged) should fix this issue as well

PR #613, opened 2026-07-23
fix(assistant): use extended markdown for all NcRichText (renders tables)

dtump:
Chat messages are rendered in markdown mode, not extended markdown, so tables aren't rendered. I'm not sure if this is what others want, but AI often outputs tables and, in my opinion, users expect those tables to render correctly. At least that's what I'd like to see, so if others agree, I propose switching to extended markdown. This complements #540 (which left tables out) and also addresses #497. Tested on NC 33 / Assistant 3.5.0. I verified that both a Markdown table and a code block (with syntax highlighting) now render correctly.

Disclosure: I used AI assistance (Claude Code) to locate the relevant prop and prepare this change; I reviewed and tested it myself.

Update (per review): dropped the now-redundant use-markdown and switched all NcRichText usages to use-extended-markdown for uniformity, so this also covers the reasoning content, the "Work with text" output, and ContextChatSource (finishing the tables part #540 left out).

Proof that after this change it correctly renders tables and code blocks:

image
image

### DIFF
--- a/src/components/ChattyLLM/Message.vue
+++ b/src/components/ChattyLLM/Message.vue
@@ -46,7 +46,7 @@
 							<div class="reasoningcontent_popover_inner">
 								<h6>{{ t('assistant', 'Reasoning content') }}</h6>
 								<NcRichText :text="message.reasoning"
-									:use-markdown="true"
+									:use-extended-markdown="true"
 									:autolink="true" />
 							</div>
 						</template>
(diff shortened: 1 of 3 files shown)


janepie:
Sounds good, let's do it! You can delete `:use-markdown="true"` in that case, it's not needed anymore.
Could you do the same change in every case we use markdown right now? Let's have some uniformity there :)

dtump:
Awesome! And thx for the feedback. I (+ Claude 😄) added the commit addressing your feedback, I think this should be sufficient?

janepie (approved):
Looks good, thanks a lot 🚀

Issue #499, opened 2026-03-25
AI summarization doesn't work for the PDFs shipping with Nextcloud ("Nextcloud Manual.pdf")


hybridherbst:
### Which version of assistant are you using?

3.3.0

### Which version of Nextcloud are you using?

v33.0.0

### Which browser are you using? In case you are using the phone App, specify the Android or iOS version and device please.

Chrome 146.0.7680.80

### Describe the Bug

Nextcloud ships with PDFs, e.g. the Nextcloud Manual.pdf. 
Nextcloud Assistant comes with "Summarize this document", so naturally I tried that on the manual.

Turns out the built-in PDF parser (smalot/pdfparser as far as I can see) fails for the built-in manual, so summarization throws an error:

<img width="566" height="260" alt="Image" src="https://github.com/user-attachments/assets/cfb6a478-1733-4431-aac6-f6f95f09745c" />

The response I see is
```json
{
    "ocs": {
        "meta": {
            "status": "failure",
            "statuscode": 400,
            "message": ""
        },
        "data": {
            "error": "Malformed UTF-8 characters, possibly incorrectly encoded"
        }
    }
}
```

and in the browser console:
```js
xhr.js:198  POST http://localhost:8080/ocs/v2.php/apps/assistant/api/v1/file-action/68/core:text2text:summary 400 (Bad request)
(anonymous) @ xhr.js:198
xhr @ xhr.js:15
xe @ dispatchRequest.js:49
_request @ Axios.js:194
request @ Axios.js:41
(anonymous) @ Axios.js:233
(anonymous) @ bind.js:12
exec @ fileActions.js:65
await in exec
At @ actionUtils.ts:48
onActionClick @ FileEntryActions.vue:187
click @ FileEntryActions.vue:9
sn @ vue.runtime.esm.js:3033
n @ vue.runtime.esm.js:1832
sn @ vue.runtime.esm.js:3033
(anonymous) @ vue.runtime.esm.js:3732
onClick @ actionText-BMig9Egt.mjs:78
handleClick @ NcActionButton-1gSJfFUC.mjs:145
sn @ vue.runtime.esm.js:3033
n @ vue.runtime.esm.js:1832
push.85471.Mr.i._wrapper @ vue.runtime.esm.js:7500Understand this error
fileActions.js:72 AxiosError: Request failed with status code 400
    at Ke (settle.js:19:12)
    at XMLHttpRequest.C (xhr.js:59:7)
[...]

PR #616, opened 2026-07-27
feat(file-action): summarize a whole file by sending it to the AI provider

cbcoutinho:
## Summary

Today the **Summarize** file action never gives the LLM the file. `TaskProcessingService::runFileAction` extracts the text with `AssistantService::parseTextFromFile` and sends that string to `core:text2text:summary`, whose `input` slot is `EShapeType::Text` — capped at **512 kB** by core.

That extraction is the weak link:

- `smalot/pdfparser` returns nothing for scanned PDFs and garbage for complex layouts (cf. #499, where the reporter's own summary was *"the PDF parser is not reliable enough for in-the-wild PDFs"*)
- PhpWord drops tables, headers and footers
- the `default:` branch returns raw bytes as "text" for unsupported binaries
- any file whose extracted text exceeds 512 kB fails validation outright, surfacing as *"Failed to launch the AI file action"*

When the configured provider can accept files, Summarize now hands over the document untouched and lets the model read it.

## How it works

A provider advertises an optional `input_attachments` (`ListOfFiles`) slot. The new `AttachmentService` feature-detects it via `getAvailableTaskTypes()[…]['optionalInputShape']`, exactly like the existing `memories` pattern, and `runFileAction` then sends:

```php
['input' => '', 'input_attachments' => [$fileId]]
```

The mandatory text input stays empty on purpose: the provider builds its own instructions around the attachment, and anything we put there would be untranslated and fight that prompt. Core accepts `''` for a mandatory Text slot, and this matches the convention already used by `MultimodalChatWithToolsProvider`.

**Nothing changes until a provider advertises the slot**, so this is inert on its own.

### Eligibility is deliberately narrow

| Type | Behaviour | Why |
|---|---|---|
| `application/pdf` | always attached | this is the actual pain point |
| `text/plain`, `text/markdown`, `text/csv` | attached only above 256 kB | extraction is lossless and cheap below that; above it, the extracted text risks blowing core's 512 kB Text cap, which fails today |
| doc / docx / odt / rtf | never attached | providers reject these outright; only our own parsers can read them |
| anything > 50 MB | never attached | mirrors the provider-side ceiling, so we fail fast rather than after a 50 MB base64 round-trip |

All three thresholds are readable from app config.

## Also included

A drive-by fix in the two file-action listeners, which both used `getFirstNodeById()` without a null check. That returns `null` when the source file is moved or deleted while the task runs — increasingly likely as tasks get longer.

In `FileActionTaskSuccessfulListener` this was worse than a plain fatal: the null deref happened *inside* the `try`, and the `catch` handler then dereferenced the same node again to send a notification. That second throw escaped the handler and aborted the whole `TaskSuccessfulEvent` dispatch, so unrelated listeners never ran.

## Depends on

nextcloud/integration_openai#419, which adds the `input_attachments` slot to `SummaryProvider` (and fixes document attachments on Mistral). This PR is safe to merge first — without a provider that advertises the slot, behaviour is byte-identical to today.

## Relationship to #602

No overlap. #602 covers chat attachments via `MultimodalChatWithTools`; this covers the Files → Summarize action, which #602 does not touch (no `TaskProcessingService`, no `fileActions.js`). The only shared surface is the initial-state listener.

## Testing

`tests/unit/Service/AttachmentServiceTest.php` — 21 tests covering the capability-detection matrix (slot absent / correct `ListOfFiles` / wrong shape type / manager throwing) and the full eligibility matrix. Passing locally.

Gates run locally: `lint`, `cs:check` (0/98), `psalm` (no errors), `composer run openapi` (`openapi.json` byte-identical — no API surface changed), `eslint`, `npm run build`.

Manual E2E still to do against a live provider: scanned PDF, `Nextcloud Manual.pdf` (#499's reproducer), a 2 MB `.md`, `.docx` still falling back to extraction, and a backend without document support falling back cleanly.

---

_This PR was generated with the help of AI, and reviewed by a Human_

### DIFF
--- a/lib/Listener/FileActionTaskFailedListener.php
+++ b/lib/Listener/FileActionTaskFailedListener.php
@@ -14,6 +14,7 @@ use OCA\Assistant\Service\TaskProcessingService;
 use OCP\EventDispatcher\Event;
 use OCP\EventDispatcher\IEventListener;
 use OCP\Files\IRootFolder;
+use OCP\Files\Node;
 use OCP\TaskProcessing\Events\TaskFailedEvent;
 use Psr\Log\LoggerInterface;
 
@@ -52,6 +53,13 @@ class FileActionTaskFailedListener implements IEventListener {
 			$this->logger->debug('FileActionTaskListener', ['source_file_id' => $sourceFileId]);
 			$userFolder = $this->rootFolder->getUserFolder($task->getUserId());
 			$sourceFile = $userFolder->getFirstNodeById($sourceFileId);
+			if (!$sourceFile instanceof Node) {
+				// the source file was removed while the task was running, there is nothing to point the user at
+				$this->logger->warning('FileActionTaskListener task failed and the source file is gone', [
+					'source_file_id' => $sourceFileId,
+				]);
+				return;
+			}
 			$this->notificationService->sendFileActionNotification(
 				$task->getUserId(), $taskTypeId, $task->getId(),
 				$sourceFileId, $sourceFile->getName(), $userFolder->getRelativePath($sourceFile->getPath()),
--- a/lib/Listener/FileActionTaskSuccessfulListener.php
+++ b/lib/Listener/FileActionTaskSuccessfulListener.php
@@ -15,6 +15,7 @@ use OCA\Assistant\Service\TaskProcessingService;
 use OCP\EventDispatcher\Event;
 use OCP\EventDispatcher\IEventListener;
 use OCP\Files\IRootFolder;
+use OCP\Files\Node;
 use OCP\SystemTag\TagNotFoundException;
 use OCP\TaskProcessing\Events\TaskSuccessfulEvent;
 use OCP\TaskProcessing\TaskTypes\TextToTextSummary;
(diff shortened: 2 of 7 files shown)


cbcoutinho:
One caveat surfaced while verifying the provider side (nextcloud/integration_openai#419) against live endpoints, because it affects what this action produces rather than whether it works.

**Some backends drop a document attachment without erroring, and the model then answers from nothing.** Tested with a 659-byte PDF containing one invented fact ("annual rainfall in Zenda is 812 millimetres"), so a wrong answer is unambiguous:

| Backend | Result (correct = 812) |
|---|---|
| Mistral | `812` ✅ |
| OpenRouter → OpenAI | `812` ✅ |
| OpenRouter → Azure | `600`, `500` ❌ silently fabricated |
| OpenRouter → Azure, with the `file-parser` plugin | `812` ✅ |

Unpinned, the same OpenRouter request alternated between routes and answers.

This is not caused by either PR — it is how those upstreams handle the (unchanged) OpenAI `type: file` part. But it is worth knowing here, because a fabricated summary is indistinguishable from a real one in `<name> - summarized.md`, whereas today's extraction path at least fails loudly.

It does not change this PR's design. The attachment path is only taken when a provider advertises `input_attachments`, so the fix belongs provider-side: either not advertising the slot for backends known to drop documents, or forcing pre-parsing. Details and a no-code-change admin workaround are in nextcloud/integration_openai#419.

---

_This comment was generated with the help of AI, and reviewed by a Human_

janepie:
Thank you for your contribution!
Please review our [AI policy](https://github.com/nextcloud/.github/blob/master/AI_POLICY.md#human-written-communication) which states that we do not want AI-generated PR summaries or comments, we want to talk to humans.
Also, we ask you to live-test this PR yourself without the help of an AI: have a Nextcloud instance running and use the FileAction with integration_openai both with and without the other PR, and ideally also with llm2 to see if there are any issues.

cbcoutinho:
Hi @janepie sure thing 👍 I'll reply once I've confirmed via testing a real server with my feedback.

Btw it looks like current developments on this app is targeting NC 35. Is there a way to backport this feature so that it works on NC 33? That's my current target instance version.

janepie:
The stable32 branch is the last one that supports NC33, but as we normally do not backport features and Assistant is changing fast these days this might not get released for NC33.

janepie:
You can use the bot for a backport though with commenting `/backport to stable32` and checking if all works as intended in the generated backport :)

PR #632, opened 2026-08-26
feat: allow the data folder name to be set server-wide

### LINKED ISSUE #631: Allow the assistant data folder name to be set server-wide
### Describe the feature you'd like

Let an administrator change the name of the folder the assistant creates in a user's files, so a deployment that does not run in English can ship a localized name to new accounts.

### Why

The assistant is presented to our users under a Turkish name, and its output lands in a folder called `Assistant` — the one English word in a file list that is otherwise entirely in their language. We would like new accounts to get that folder under the name they already know the feature by.

We run a Turkish-language deployment with several thousand accounts.

### What is possible today

Half of this already works, which is why the gap is a little frustrating. `AssistantService::getAssistantDataFolder()` reads a per-user value:

```php
$dataFolderName = $this->config->getUserValue($userId, Application::APP_ID, 'data_folder', Application::ASSISTANT_DATA_FOLDER_NAME) ?: Application::ASSISTANT_DATA_FOLDER_NAME;
```

So an existing folder can be pointed at per user. What is missing is a default behind it: `Application::ASSISTANT_DATA_FOLDER_NAME` is a plain constant, so every new account gets `Assistant` and there is no app config for an administrator to set.

### A related inconsistency

While looking at this we noticed that setting `data_folder` to a name that does not exist yet does not do what it looks like it does. `getAssistantDataFolder()` falls through to `createAssistantDataFolder()` when the configured folder is missing, and that method ignores the configured name entirely:

```php
private function createAssistantDataFolder(string $userId, int $try = 0): Folder {
	$userFolder = $this->rootFolder->getUserFolder($userId);
	if ($try === 0) {
		$folderPath = Application::ASSISTANT_DATA_FOLDER_NAME;
	} else {
		$folderPath = Application::ASSISTANT_DATA_FOLDER_NAME . ' ' . $try;
	}
	...
}
```

So an administrator who runs `occ user:setting <uid> assistant data_folder Bilge8` for a user who has never used the assistant gets a folder called `Assistant`, and the caller then writes that name back over the preference:

```php
$dataFolder = $this->createAssistantDataFolder($userId);
$dataFolderName = $dataFolder->getName();
$this->config->setUserValue($userId, Application::APP_ID, 'data_folder', $dataFolderName);
```

The setting silently reverts. Whether or not the app config below is wanted, having `createAssistantDataFolder()` honour the configured name would make the existing per-user setting behave the way it reads.

### Prior art in the same ecosystem

Talk solves the same problem in `spreed/lib/Config.php`:

```php
$defaultAttachmentFolder = $this->config->getAppValue('spreed', 'default_attachment_folder', '/Talk');
return $this->config->getUserValue($userId, 'spreed', UserPreference::ATTACHMENT_FOLDER, $defaultAttachmentFolder);
```

An app-level default behind the per-user value, so `occ config:app:set spreed default_attachment_folder` changes what new users get without touching anyone who already has a folder.

### Suggested implementation

Resolve the default once and pass it down, so both the lookup and the creation use the same name:

```diff
 	public function getAssistantDataFolder(string $userId): Folder {
 		$userFolder = $this->rootFolder->getUserFolder($userId);
 
-		$dataFolderName = $this->config->getUserValue($userId, Application::APP_ID, 'data_folder', Application::ASSISTANT_DATA_FOLDER_NAME) ?: Application::ASSISTANT_DATA_FOLDER_NAME;
+		$defaultName = $this->config->getAppValue(Application::APP_ID, 'default_data_folder', Application::ASSISTANT_DATA_FOLDER_NAME) ?: Application::ASSISTANT_DATA_FOLDER_NAME;
+		$dataFolderName = $this->config->getUserValue($userId, Application::APP_ID, 'data_folder', $defaultName) ?: $defaultName;
 		if ($userFolder->nodeExists($dataFolderName)) {
 			...
 		}
 		// it does not exist or is not a folder or does not have write permissions: we create one
-		$dataFolder = $this->createAssistantDataFolder($userId);
+		$dataFolder = $this->createAssistantDataFolder($userId, $dataFolderName);
```

with `createAssistantDataFolder()` taking the base name instead of reading the constant, keeping its existing " 1", " 2" collision suffixes.

Nothing changes for a server that sets no app config, existing folders keep working because the per-user value still wins, and the per-user setting starts behaving as written.

I am happy to open a pull request with this if the approach looks right to you.

bakiburakogun:
Fixes #631

### Summary

`Application::ASSISTANT_DATA_FOLDER_NAME` is a class constant, so a server whose users do not work in English has no supported way to give new accounts a localized name for the folder the assistant creates in their files.

The per-user `data_folder` setting does not fill the gap. When the configured folder does not exist yet, `getAssistantDataFolder()` falls through to `createAssistantDataFolder()`, which reads the constant rather than the configured name, and the caller then writes the created name back over the preference:

```php
$dataFolder = $this->createAssistantDataFolder($userId);
$dataFolderName = $dataFolder->getName();
$this->config->setUserValue($userId, Application::APP_ID, 'data_folder', $dataFolderName);
```

So `occ user:setting <uid> assistant data_folder Something` on a user who has not used the assistant yet creates `Assistant` and silently reverts the setting.

### What this changes

Resolve the default from an app config value, keeping the constant as the last resort, and pass the resolved name down so the creation path uses it:

```php
$defaultFolderName = $this->config->getAppValue(Application::APP_ID, 'default_data_folder', Application::ASSISTANT_DATA_FOLDER_NAME) ?: Application::ASSISTANT_DATA_FOLDER_NAME;
$dataFolderName = $this->config->getUserValue($userId, Application::APP_ID, 'data_folder', $defaultFolderName) ?: $defaultFolderName;
```

`createAssistantDataFolder()` now takes the base name instead of reading the constant, keeping its existing " 1", " 2" collision suffixes.

An administrator can then run:

```
occ config:app:set assistant default_data_folder --value="Bilge8"
```

This is the same shape Talk uses for its attachment folder (`default_attachment_folder` in `spreed/lib/Config.php`): an app-level default behind the per-user value.

### Behaviour

- A server that sets no app config behaves exactly as before.
- Users who already have a folder keep it, because the per-user value still wins and is written on first use.
- The per-user setting now works for a folder that does not exist yet, which it did not before.

### Testing

Running on Nextcloud 34 with the equivalent change against the bundled version, on a server with several thousand accounts. New accounts get the configured folder name, existing accounts keep theirs, and unsetting the app config restores the previous behaviour.

`php-cs-fixer` and `psalm` are clean for the changed file.

### DIFF
--- a/lib/Service/AssistantService.php
+++ b/lib/Service/AssistantService.php
@@ -24,6 +24,7 @@ use OCP\Files\GenericFileException;
 use OCP\Files\InvalidPathException;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotPermittedException;
+use OCP\IAppConfig;
 use OCP\IConfig;
 use OCP\IL10N;
 use OCP\ITempManager;
@@ -91,6 +92,7 @@ class AssistantService {
 		private IL10N $l10n,
 		private ITempManager $tempManager,
 		private IConfig $config,
+		private IAppConfig $appConfig,
 		private IShareManager $shareManager,
 		private SystemTagService $systemTagService,
 	) {
@@ -522,7 +524,7 @@ class AssistantService {
 	public function getAssistantDataFolder(string $userId): Folder {
 		$userFolder = $this->rootFolder->getUserFolder($userId);
 
-		$defaultFolderName = $this->config->getAppValue(Application::APP_ID, 'default_data_folder', Application::ASSISTANT_DATA_FOLDER_NAME) ?: Application::ASSISTANT_DATA_FOLDER_NAME;
+		$defaultFolderName = $this->appConfig->getValueString(Application::APP_ID, 'default_data_folder', Application::ASSISTANT_DATA_FOLDER_NAME, lazy: true) ?: Application::ASSISTANT_DATA_FOLDER_NAME;
 		$dataFolderName = $this->config->getUserValue($userId, Application::APP_ID, 'data_folder', $defaultFolderName) ?: $defaultFolderName;
 		if ($userFolder->nodeExists($dataFolderName)) {
 			$dataFolderNode = $userFolder->get($dataFolderName);
@@ -531,7 +533,7 @@ class AssistantService {
 			}
 		}
 		// it does not exist or is not a folder or does not have write permissions: we create one
-		$dataFolder = $this->createAssistantDataFolder($userId, $defaultFolderName);
+		$dataFolder = $this->createAssistantDataFolder($userId, $dataFolderName);
 		$dataFolderName = $dataFolder->getName();
 		$this->config->setUserValue($userId, Application::APP_ID, 'data_folder', $dataFolderName);
 		return $dataFolder;
--- a/lib/Service/AssistantService.php
+++ b/lib/Service/AssistantService.php
@@ -522,7 +522,8 @@ class AssistantService {
 	public function getAssistantDataFolder(string $userId): Folder {
 		$userFolder = $this->rootFolder->getUserFolder($userId);
 
-		$dataFolderName = $this->config->getUserValue($userId, Application::APP_ID, 'data_folder', Application::ASSISTANT_DATA_FOLDER_NAME) ?: Application::ASSISTANT_DATA_FOLDER_NAME;
+		$defaultFolderName = $this->config->getAppValue(Application::APP_ID, 'default_data_folder', Application::ASSISTANT_DATA_FOLDER_NAME) ?: Application::ASSISTANT_DATA_FOLDER_NAME;
+		$dataFolderName = $this->config->getUserValue($userId, Application::APP_ID, 'data_folder', $defaultFolderName) ?: $defaultFolderName;
 		if ($userFolder->nodeExists($dataFolderName)) {
 			$dataFolderNode = $userFolder->get($dataFolderName);
 			if ($dataFolderNode instanceof Folder && $dataFolderNode->isCreatable()) {
@@ -530,7 +531,7 @@ class AssistantService {
 			}
 		}
 		// it does not exist or is not a folder or does not have write permissions: we create one
-		$dataFolder = $this->createAssistantDataFolder($userId);
+		$dataFolder = $this->createAssistantDataFolder($userId, $defaultFolderName);
 		$dataFolderName = $dataFolder->getName();
 		$this->config->setUserValue($userId, Application::APP_ID, 'data_folder', $dataFolderName);
 		return $dataFolder;
@@ -538,17 +539,18 @@ class AssistantService {
 
 	/**
 	 * @param string $userId
+	 * @param string $baseName
 	 * @param int $try
 	 * @return Folder
 	 * @throws NoUserException
 	 * @throws NotPermittedException
 	 */
-	private function createAssistantDataFolder(string $userId, int $try = 0): Folder {
+	private function createAssistantDataFolder(string $userId, string $baseName, int $try = 0): Folder {
 		$userFolder = $this->rootFolder->getUserFolder($userId);
 		if ($try === 0) {
-			$folderPath = Application::ASSISTANT_DATA_FOLDER_NAME;
+			$folderPath = $baseName;
 		} else {
-			$folderPath = Application::ASSISTANT_DATA_FOLDER_NAME . ' ' . $try;
+			$folderPath = $baseName . ' ' . $try;
 		}
 
 		if ($userFolder->nodeExists($folderPath)) {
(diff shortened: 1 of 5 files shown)


lib/Service/AssistantService.php
```diff
@@ -522,41 +522,43 @@ public function storeInputFile(string $userId, string $tempFileLocation, ?string
...
+		$dataFolderName = $this->config->getUserValue($userId, Application::APP_ID, 'data_folder', $defaultFolderName) ?: $defaultFolderName;
 		if ($userFolder->nodeExists($dataFolderName)) {
 			$dataFolderNode = $userFolder->get($dataFolderName);
 			if ($dataFolderNode instanceof Folder && $dataFolderNode->isCreatable()) {
 				return $dataFolderNode;
 			}
 		}
 		// it does not exist or is not a folder or does not have write permissions: we create one
-		$dataFolder = $this->createAssistantDataFolder($userId);
+		$dataFolder = $this->createAssistantDataFolder($userId, $defaultFolderName);
```
julien-nc:
```suggestion
		$dataFolder = $this->createAssistantDataFolder($userId, $dataFolderName);
```
The default one is used only if the user didn't define a value for `data_folder`. With the current line, only the default value is used and the user value is never used.

julien-nc (changes requested):
Looks good. This would need to be documented at some point or even better, allow to change those values in the admin settings (the default) and user settings (the user value).

bakiburakogun:
Thanks both — pushed two commits.

**@coderabbitai** was right, and it was the same class of bug the PR set out to fix, just one line further down. `createAssistantDataFolder()` was passed `$defaultFolderName` instead of the resolved `$dataFolderName`, so a user whose configured folder had been deleted would get one named after the administrator's default, and their own setting was then overwritten with it. Fixed in c8c9b06.

**@julien-nc** — added both settings rather than only documenting it, since that was the option you preferred:

- Admin settings get a "Data folder" section with the server-wide default (`default_data_folder`).
- Personal settings get the per-user override (`data_folder`), with the administrator's default shown as the placeholder. Leaving it empty falls back to the default, so the precedence is visible in the interface rather than buried in the code.

Both use the existing config endpoints, so no new routes. I moved the app-config read to `IAppConfig::getValueString(..., lazy: true)` to match how `setAdminConfig()` writes and how the rest of the app reads its lazy values.

Two things I did not do, happy to add either if you want them:

- The personal field accepts any string, the same as the existing `occ user:setting` path did. I did not add validation for path separators or reserved names, since `createAssistantDataFolder()` already falls back safely when the name cannot be created, but say the word if you would rather it be restricted.
- No admin documentation page, as I could not find an existing one in this repo to extend.

`php-cs-fixer`, `psalm` and `eslint` are clean, and the frontend builds.

bakiburakogun:
Both fixed in 055bb42.

**The `"0"` case** — right, and it was in three places rather than one: the service and both settings pages all used `?:`, which PHP treats `"0"` as false for. A folder actually named `0` would have been replaced by the built-in name. Now checked with `=== ''` at all three sites.

**The delayed save** — also right, and worth spelling out because it is not specific to this PR: `delay()` in `src/utils.js` keeps one module-level timer and clears it on every call, so editing a second field within the two seconds cancelled the first callback before it had ever written into `optionsToSave`. That edit was simply lost. The value is now queued when it changes rather than when the timer fires, and `saveOptions` gets a snapshot so a later edit cannot mutate the set mid-request.

That one is pre-existing behaviour in the admin settings rather than something this PR introduced — I fixed it here because the new field is subject to it and the fix is four lines. @julien-nc, if you would rather have it as a separate PR, say so and I will split it out.

Checks are clean: `php-cs-fixer`, `psalm`, `eslint`, and the frontend builds.

bakiburakogun:
Fixed in d3dcaa7 — and this one was worth catching, because the code disagreed with what the admin settings text promises.

A user who had used the assistant before an administrator set a default can have a folder under the built-in name with no stored `data_folder` value. The server default was being applied to them, which started a second folder and left their existing output behind it, while the settings text says changing the default does not touch folders that already exist.

Now an empty per-user value falls back to the built-in name when that folder already exists, and to the server default only when there is nothing to strand:

```php
$dataFolderName = $this->config->getUserValue($userId, Application::APP_ID, 'data_folder', '');
if ($dataFolderName === '') {
	$dataFolderName = $userFolder->nodeExists(Application::ASSISTANT_DATA_FOLDER_NAME)
		? Application::ASSISTANT_DATA_FOLDER_NAME
		: $defaultFolderName;
}
```

I did not add a migration that renames existing folders. Moving a user's files because an administrator changed a setting seems like more than this PR should do on its own, and the existing behaviour of leaving them in place is at least predictable. Happy to add one if you would rather the rename be offered explicitly.

DCO was also failing on my earlier commits — sorry, that was mine, the sign-off was missing on three of them. Rebased with `--signoff`; the diff is unchanged. Both checks are green now.

`php-cs-fixer`, `psalm` and `eslint` are clean and the frontend builds.

julien-nc (approved):
Thanks a lot.

> if you would rather have it as a separate PR, say so and I will split it out.

It's fine here, the change is kind of related.
Tested, works great.

